PR #103, opened 2025-08-24
67 bug goals removed from the canvas cannot be reused later

ToriLi1210:


src/components/SectionPanel.tsx
```diff
@@ -126,6 +126,18 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
   const goalListRef = useRef<HTMLDivElement>(null);
   const timeoutRef = useRef<NodeJS.Timeout | null>(null);
 
+  // define a handler for deleting a cell and it children from the Graph Render Section
+  const handleDeleteCellsFromGraph = (treeItems:TreeItem[]) => {
+  // remove goal from treeData / cluster hierarchy
+    dispatch(removeItemIdFromTree);
+  };
+
```
bggolding, line 134: I think the `removeItemFromTree` function needs an argument, in fact, the body should probably be:
`treeItems.forEach((treeItem) => dispatch(removeItemIdFromTree(treeItem.id)));`

src/components/context/treeDataSlice.ts
```diff
@@ -105,11 +116,19 @@ export const treeDataSlice = createSlice({
             state.tree.push(newTreeNode({goalId: action.payload.id}));
             state.treeIds.push(action.payload.id);
         },
+        // remove goal(s) and its children 
+        removeGoalToTree: (state, action: PayloadAction<{
```
bggolding: I think it would read better as `removeGoalFromTree`

src/components/context/treeDataSlice.ts
```diff
@@ -41,17 +41,28 @@ const createGoalsAndTabsFromTabContent = (initialTabs: InitialTab[]): {
...
+export const removeItemIdFromTree = (
+  items: TreeNode[],
+  id: TreeNode["goalId"],
+  removeChildren: boolean=true
+): TreeNode[] => {
+  return items.reduce((acc, item) => {
+    if (item.goalId === id) {
+      if (!removeChildren && item.children) {
+        // Promote children to parent level
+        acc.push(...item.children);
```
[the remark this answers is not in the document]
ToriLi1210: Thanks Ben! You’re absolutely right. We don’t need to promote children because when removeChildren is set to true, deleting the parent will automatically delete the entire subtree (all its children and descendants). So there’s no need to promote the children of those child nodes upward.

Issue #102, opened 2025-08-21
Bug: Export button autofocus issue after exporting


lvqq2000:
Steps to reproduce:
1. Export a non-empty graph.
2. Return to the application and click on one goal.
3. ress Del to attempt deletion.

Expected result
The selected goal should be deleted.

Actual result
Focus is automatically set to the Export button. The goal is not deleted. The user must click the goal again before performing the deletion.

aoxiangx:
A new problem appeared after fix:

Unable to delete the goal after re-enter the canvas through back button.

Steps to reproduce:

1. Click `Create Model` from home page.
2. Click `back` button to return home page.
3. Select any goals and try Delete through keyboard


Expected result
The selected goal should be deleted.

Actual result
Nothing happened.

PR #123, opened 2025-09-15
#102 export button autofocus

aoxiangx:
fix the autofocus issue and some other minor issues(mostly warnings)

### DIFF
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -14,13 +14,11 @@ import {
 } from "@maxgraph/core";
 import '@maxgraph/core/css/common.css';
 
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { Button, Col, Container, Form, Row } from "react-bootstrap";
+import {useCallback, useEffect, useMemo, useRef, useState} from "react";
+import {Col, Container, Form, Row} from "react-bootstrap";
 import ErrorModal, { ErrorModalProps } from "../ErrorModal.tsx";
-import { associateNonFunctions, layoutFunctions, renderGoals } from './GraphHelpers';
-import {
-  registerCustomShapes,
-} from "./GraphShapes";
+import {associateNonFunctions, layoutFunctions, renderGoals} from './GraphHelpers';
+import {registerCustomShapes} from "./GraphShapes";
 import "./GraphWorker.css";
 import {useFileContext} from "../context/FileProvider.tsx";
 import {useGraph} from "../context/GraphContext";
@@ -49,7 +47,6 @@ interface CellHistory {
 }
 
 interface GlobObject {
-  [key: string]: string[];
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/components/header/ExportFileButton.tsx
```diff
@@ -127,6 +127,14 @@ const ExportFileButton = ({ showGraphSection }: { showGraphSection: boolean }) =
 		catch (error) {
 			console.error('Failed to save file: ', error);
 		}
+
+		// Return focus to graph container to enable keyboard shortcuts
+		setTimeout(() => {
```
bggolding, line 136: I'm really not happy with this fix. It's not clear why sleeping 100ms is an actual fix for this problem, it looks just like a workaround, a hack that happens to work. This sort of wishful programming rapidly makes an application hard to maintain because noone can be sure why these bits of magic work and they dare not touch them. It also just makes the app slower because there are all these sleeps happening.

We need to understand what the actual problem is. Is there something that's "settling" during the rendering that makes this work? If so, is there an event that's raised or a callback that can be called that will alert us when the task has completed so we can continue as soon as that happens? Is it a bug in this version of the library? Did you look to see if there's a later version than the one we're using?

You've repeated them same code in several places making it difficult to update or use somewhere else. The DRY principle (Don't Repeat Yourself) is something to really try to hold to.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -47,6 +47,7 @@ interface CellHistory {
 }
 
 interface GlobObject {
+    [key: string]: { id: number; content: string; }[];
```
bggolding, line 50: Why don't you use the definition that's in `components/types.ts` which is the same as this? Again, Don't Repeat Yourself.

Can we try to format this better? 
```typescript
interface GlobObject {
    [key: string]: {
        id: number;
        content: string;
    }[];
}
```

src/components/header/ExportFileButton.tsx
```diff
@@ -194,8 +201,13 @@ const ExportFileButton = ({ showGraphSection }: { showGraphSection: boolean }) =
 		// Use Canvg to render SVG onto the canvas
 		const v = Canvg.fromString(context, svgString);
 
-		// Render SVG onto the canvas
-		await v.render();
+		// Render SVG onto the canvas with error handling
+		try {
+			await v.render();
+		} catch (imageError) {
+			console.warn('Image loading error during PNG export:', imageError);
```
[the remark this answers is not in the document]
bggolding: You could even write this as:
```typescript
    v.render()
        .catch((imageError) => {
			console.warn('Image loading error during PNG export:', imageError);
        });
```

src/components/header/ExportFileButton.tsx
```diff
@@ -89,7 +89,11 @@ const ExportFileButton = ({ showGraphSection }: { showGraphSection: boolean }) =
 	
 		// Serialize the SVG element to a string
 		const serializer = new XMLSerializer();
-		const svgString = serializer.serializeToString(svgElement);
+		let svgString = serializer.serializeToString(svgElement);
```
bggolding, line 92: I would really try not to use `let` if at all possible. The fact that you've had to use a comment is an indicator that a well-named function could be used here. Say:
```typescript
const svgString = sanitiseSvgString(serializer.serializeToString(svgElement));
```
You can expand the sanitisation of the string further there as needed.

Again, you have repeated code. DRY. This should be a single routine to do this so that it can be tested and reused.

bggolding:
Please rework this.

bggolding (changes requested):
We can talk about this in the meeting today.

Issue #122, opened 2025-09-15
Unable to delete the goal after re-enter the canvas through back button


aoxiangx:
Steps to reproduce:

1. Click Create Model from home page, click render.
2. Click Home button to return home page.
3. Click Create Model, click render,  select any goals and try Delete through keyboard

Expected result:
The selected goal should be deleted.

Actual result:
Nothing happened.

aoxiangx:
close as PR is passed

PR #142, opened 2025-09-23
#122 Unable to delete the goal after re-enter the canvas through back button / re-render

aoxiangx:
fix(graph container): fix the problem that the container will lose focus after navigation/re-render

#122

### DIFF
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -16,11 +16,9 @@ import '@maxgraph/core/css/common.css';
 
 import {useCallback, useEffect, useMemo, useRef, useState} from "react";
 import {Button, Col, Container, Form, Row} from "react-bootstrap";
-import ErrorModal, { ErrorModalProps } from "../ErrorModal.tsx";
+import ErrorModal, {ErrorModalProps} from "../ErrorModal.tsx";
 import {associateNonFunctions, isGoalNameEmpty, layoutFunctions, renderGoals} from './GraphHelpers';
-import {
-  registerCustomShapes,
-} from "./GraphShapes";
+import {registerCustomShapes} from "./GraphShapes";
 import "./GraphWorker.css";
 import {useFileContext} from "../context/FileProvider.tsx";
 import {useGraph} from "../context/GraphContext";
@@ -83,7 +81,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
   const cells = deletingItemRef.current
   if(!cells||!graph) return
   const deletedCells: Cell[] = [];
-  
+
   // selected cell
   const removeCellRecursively = (cell: Cell) => {
     // check the children
@@ -98,7 +96,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
     deletedCells.push(...removed);
   };
 
-  cells.forEach(cell => 
+  cells.forEach(cell =>
       removeCellRecursively(cell));
   deletedCells.forEach((cell) => {
     // since the cell.getID is functionatype + the id
@@ -124,7 +122,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
   //     onResetDefault();
   //   }
   // };
-  
+
   // Track if we have already centered on first entry
   const hasCenteredOnEntryRef = useRef(false);
   const prevShowGraphSectionRef = useRef(false);
@@ -137,7 +135,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
       graph.center();
     }
   };
-  
+
   const initRecentreView = useCallback(() => {
     if (graph) {
       graph.fit();
@@ -330,13 +328,13 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
     const keyHandler = new KeyHandler(graph);
     keyHandler.bindKey(DELETE_KEYBINDING, () => {
       if (graph.isEnabled()) {
-        
+
         const selectedCells = graph.getSelectionCells();
         if (!selectedCells || selectedCells.length === 0) return;
 
         deletingItemRef.current = selectedCells;
 
-      
+
 
         const outgoingEdges = graph.getOutgoingEdges(selectedCells[0],null);
         const hasChildren = outgoingEdges.some(edge => edge.target && edge.target !== selectedCells[0]);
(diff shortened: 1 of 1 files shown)


bggolding (changes requested):
Does this need to be in a useEffect at all? If either `graph` or `cluster` change, that will trigger a re-render of this component during which `renderGraph` and `showGraphSection` will be recreated with new closures.

There are a lot of unnecessary useEffect uses in this app which do nothing more than slow it down by triggering unnecessary re-renders. We need to look out for them and try to remove them when we can. It makes the code much clearer and easier to read and reason about.

It looks like some of the earlier developers did not understand how React triggers a re-render of a component, ie, that happens when a parameter changes or a value changes from a hook.

Please turn off auto formatting. We are using 4 spaces indents in this project, not 2. I don't want spaces inside braces.

Your comment before the `returnFocusToGraph()` is redundant -- the function has an excellent name that makes it entirely clear what its purpose is.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -535,20 +543,16 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
...
-        renderGraph();
-      } 
-      else {
-        graph.getDataModel().clear();
-        console.log("Graph is empty");
-      }
+  // Handle graph rendering when cluster or graph changes
+  if (graph) {
+    if (cluster.ClusterGoals.length > 0) {
+      console.log("re render");
```
bggolding, line 549: Please don't commit `console.log` statements.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -566,6 +570,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
       // Use setTimeout to ensure centering happens after layout is complete
       setTimeout(() => {
         initRecentreView();
+        returnFocusToGraph();
```
bggolding, line 573: I am still really unhappy about using a timeout to trigger this action. How do you know that 200ms is the correct amount of time to wait for the rendering to settle? This sort of change makes it very difficult for someone else to understand and modify this code later because they don't know how this constant was decided on, why it was necessary and what other changes to the code will affect it.

Is there an event or callback that's triggered by the rendering library that you can use to be alerted when the rendering process has completed? Is there something in React that you can hook into at the end of the current render cycle and trigger based on that?

src/components/Graphs/GraphWorker.tsx
```diff
@@ -500,7 +501,14 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
       stakeholdersGlob
     );
     graph.getDataModel().endUpdate();
-  }, [graph, cluster, initRecentreView]);
+
+    if (showGraphSection) {
+      // Use setTimeout to ensure focus is set after DOM updates
+      setTimeout(() => {
+        returnFocusToGraph();
```
bggolding, line 508: If the timeout on the call is 0s, why don't you just call the function directly? Is it to push the call to the next render cycle? This looks like a clumsy way to do this.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -31,6 +31,7 @@ import WarningMessage from "./WarningMessage";
 import {VERTEX_FONT} from "../utils/GraphConstants.tsx"
 import {removeGoalIdFromTree} from "../context/treeDataSlice.ts";
 import ConfirmModal from "../ConfirmModal.tsx";
+import { returnFocusToGraph } from "../utils/GraphUtils.tsx";
```
bggolding, line 34: Please remove spaces inside braces.

aoxiangx:
For the timeout thing, I tried to remove it but the fix just suddenly not working anymore. I might need more time to look into this issue.

src/components/Graphs/GraphWorker.tsx
aoxiangx: Initially, I used `setTimeout(..., 0`) because I thought this was a timing issue. The magic number 0 forces the browser to defer execution to the end of the render stack, which seemed to solve the problem by reordering when the focus action was triggered.

However, after further investigation, I discovered the root cause: MaxGraph performs internal asynchronous DOM operations after endUpdate() completes. These operations interfere with focus management, causing our focus restoration to be overridden.

**Attempted Solutions:**

1. Direct `returnFocusToGraph()` without timeout - Didn't work because MaxGraph's DOM updates took control again afterward
2. `useLayoutEffect` - Executed too early, before MaxGraph's internal operations finished
3. Listener on` END_UPDATE` event - Only signals the end of the event, not the completion of canvas rendering

**Why setTimeout(fn, 0) worked:** It forced the call to be delayed until the next event loop tick, ensuring all synchronous and asynchronous operations within MaxGraph were completed first.

**Final Solution:**
Add a listener to the Click event instead. This approach is more logical because:

a) Users must click a goal to select it before they can delete it via keyboard
b) Restoring focus on every goal click is intuitive and easy to understand

This solution is cleaner and more maintainable than relying on timing workarounds.

src/components/Graphs/GraphWorker.tsx
aoxiangx: fixed in the lastest commit

bggolding (changes requested):
I am very impressed with your solution to this Aoxaing. Great work!

The comments here are minor and just try to make the code fit into the style of the app so that it looks consistent.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -404,6 +404,17 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
...
   };
 
+    // Ensure mouse interactions restore focus so keyboard shortcuts (e.g., Delete) work reliably
+    if(graph){
+        graph.addListener(InternalEvent.CLICK, (_sender: string, _evt: EventObject) => {
+            returnFocusToGraph();
+            graph.refresh();
+        });
+        graph.getSelectionModel().addListener(InternalEvent.CHANGE, (_sender: string, _evt: EventObject) => {
+        returnFocusToGraph();
```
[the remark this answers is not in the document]
aoxiangx: Yes, I'm sorry that I didn't clean it up from my previous attempt, fixed it.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -404,6 +404,17 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
     };
   };
 
+    // Ensure mouse interactions restore focus so keyboard shortcuts (e.g., Delete) work reliably
+    if(graph){
```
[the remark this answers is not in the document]
aoxiangx: Got it.

Issue #106, opened 2025-08-27
Bug:Empty input handling & Character Goal edit problems in Canvas


ToriLi1210:
Description:
There are two editing issues in the graph editor:

1. Empty input handling

When the user edits a node and leaves the input empty, the graph accepts it.

Expected behavior: Show an alert (e.g., "Input cannot be empty") and revert the change.

2. Character Goal editing

When editing a Character Goal:

If there is more than one element, edits do not update correctly.

lollyluan:
Hi @leonsterling,
Currently, when editing the text of Non-Functional Goals in the canvas, the logic works as follows:

- It compares the number of goals (split by commas) in the canvas text with the number of Non-Functional Goals in the hierarchy.

- If the numbers do not match, an alert is shown.

Would you like to keep this behavior, or would you prefer that when a user edits the canvas text (and splits by commas), it should automatically update the hierarchy instead?

PR #147, opened 2025-09-24
#106 Adjust non-functional node sizing and empty input handling

lollyluan:
1. Show an alert when the user leaves a Functional Goal empty. 

> (Note: further discussion may be needed on whether to block empty input or allow it with warnings).

2. Adjust Non-Functional Goal context label sizing. Labels are now displayed as squares, with height and width adjusted for balance.


bggolding (changes requested):
I think `Input` and `newInput` aren't very good names for what is a goal name. The message in the error modal should be changed to reflect that too.

The title in the error message is okay, it's the message that needs to direct the user to the part of the input that's a problem, ie, "Goal name cannot be empty".

I would replace the comment and test with a small function:

```typescript
if (isGoalNameEmpty(change.value)) {
    ...
}
```

There is a function called `isGoalEmpty` in `GoalHint.tsx` but I'm not sure that we can use that from here.

lollyluan:
> I think `Input` and `newInput` aren't very good names for what is a goal name. The message in the error modal should be changed to reflect that too.
> 
> The title in the error message is okay, it's the message that needs to direct the user to the part of the input that's a problem, ie, "Goal name cannot be empty".
> 
> I would replace the comment and test with a small function:
> 
> ```ts
> if (isGoalNameEmpty(change.value)) {
>     ...
> }
> ```
> 
> There is a function called `isGoalEmpty` in `GoalHint.tsx` but I'm not sure that we can use that from here.

I checked the existing `isGoalEmpty` in `GoalHint.tsx`, but that function expects a full goal object as its parameter.
In this case, we only have access to the raw input value as a string, so we can’t reuse it directly.

I wrote a small helper `isGoalNameEmpty(value: string)` here. It only checks whether the goal name string is empty or whitespace, which fits our use case.

bggolding (approved):
That looks really good. Thanks.

PR #148, opened 2025-09-28
146 dragging goal to cluster creates duplicate copies and inherits previous children

ToriLi1210:
Description:

This PR addresses an issue where dragging treeitem with its children caused duplicated nodes.

Problem:

Each tree item in Nestable previously default used id as its primary key.

Since the same goal can be added multiple times to the tree (as long as it is not in the same hierarchy), the id alone is no longer unique.

This caused Nestable to clone nodes multiple times and triggered React warnings about non-unique keys.

Solution:

Introduce an extra property instanceID for each TreeItem. The format is "id-instanceID".

Use instanceID as the primary key for Nestable by setting the idProp="instanceID" on the Nestable component.

This ensures that each item in the tree has a unique key, preventing duplicate cloning while allowing multiple instances of the same goal.

### DIFF
--- a/src/components/Tree.tsx
+++ b/src/components/Tree.tsx
@@ -264,6 +264,7 @@ const Tree: React.FC<TreeProps> = ({
             ? "#FF474C"
             : "white",
         }}
+        key={`${treeItem.id}-${treeItem.instanceID}`}
         className="tree-list"
         onDoubleClick={handleDoubleClick}
       >
--- a/src/components/Tree.tsx
+++ b/src/components/Tree.tsx
@@ -144,7 +144,7 @@ const Tree: React.FC<TreeProps> = ({
   //   return ids;
   // };
 
-  const getAllGoalInstances = (item: TreeItem): { goalId: number; instanceID: number }[] => {
+  const getAllGoalInstances = (item: TreeItem): { goalId: TreeItem["id"]; instanceID: TreeItem["instanceID"] }[] => {
     const result = [{ goalId: item.id, instanceID: item.instanceID }];
 
     if (item.children) {
@@ -360,6 +360,7 @@ const Tree: React.FC<TreeProps> = ({
         onChange={({ items }) => dispatch(setTreeData(items as TreeItem[]))}
         items={treeData}
         renderItem={renderItem}
+        idProp="instanceID"
         renderCollapseIcon={({ isCollapsed }) => (
           <Collapser isCollapsed={isCollapsed} />
         )}
--- a/src/components/Tree.tsx
+++ b/src/components/Tree.tsx
@@ -97,7 +97,7 @@ const Tree: React.FC<TreeProps> = ({
   // Delete item by its id
   const deleteItem = () => {
     if (deletingItemRef?.current) {
-      dispatch(deleteGoal(deletingItemRef.current));
+      dispatch(deleteGoal({item:deletingItemRef.current,deleteFromGoalList:false}));
     }
     setShowDeleteWarning(false);
   };
--- a/src/components/Tree.tsx
+++ b/src/components/Tree.tsx
@@ -20,7 +20,7 @@ import {
 } from "../components/utils/GoalHint.tsx"
 
 import "./Tree.css";
-import {deleteGoal, setTreeData} from "./context/treeDataSlice.ts";
+import {deleteGoalReferenceFromHierarchy, setTreeData} from "./context/treeDataSlice.ts";
 
 // Inline style for element in Nestable, css style import not working
 const treeListStyle: React.CSSProperties = {
@@ -97,7 +97,7 @@ const Tree: React.FC<TreeProps> = ({
   // Delete item by its id
   const deleteItem = () => {
     if (deletingItemRef?.current) {
-      dispatch(deleteGoal({item:deletingItemRef.current,deleteFromGoalList:false}));
+      dispatch(deleteGoalReferenceFromHierarchy({item:deletingItemRef.current}));
     }
     setShowDeleteWarning(false);
   };
--- a/src/components/Tree.tsx
+++ b/src/components/Tree.tsx
@@ -61,9 +61,11 @@ const iconFromType = (type: Label) => {
 };
 
 type TreeProps = {
-  existingItemIds: number[];
+  // existingItemIds: number[];
   handleSynTableTree: (treeItem: TreeItem, editedText: string) => void;
-  setExistingItemIds: (existingItemIds: number[]) => void;
+  // setExistingItemIds: (existingItemIds: number[]) => void;
+  existingGoalReferenceInstanceId: { goalId: TreeItem["id"]; instanceID: TreeItem["instanceID"] }[];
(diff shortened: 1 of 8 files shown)


src/components/Tree.tsx
```diff
@@ -144,7 +144,7 @@ const Tree: React.FC<TreeProps> = ({
   //   return ids;
   // };
 
-  const getAllGoalInstances = (item: TreeItem): { goalId: number; instanceID: number }[] => {
+  const getAllGoalInstances = (item: TreeItem): { goalId: TreeItem["id"]; instanceID: TreeItem["instanceID"] }[] => {
```
bggolding, line 147: Excellent! Updating the loose `number` type to the specific type `TreeItem["id"]` is the sort of small change that incrementally improves the quality of the codebase.

The conflict in names between `goalId` and `instanceID` is a glaring wart and should be addressed. I will open a ticket to change that name to make it consistent. These sorts of arbitrary name discrepancies contribute to make extra mental load on developers.

src/components/context/FileProvider.tsx
```diff
@@ -26,12 +26,16 @@ export interface TreeNode {
 
 // require id and type fields, others optional.
 // create a empty treeItem
-export const newTreeItem = (initFields: Pick<TreeItem, "type"> & Partial<TreeItem>): TreeItem => ({
-    id: initFields.id ?? Date.now(),
-    content: "",
-    instanceID:initFields.instanceID??0,  // give 0 when it is empty
-    ...initFields
-});
+export const newTreeItem = (initFields: Pick<TreeItem, "type"> & Partial<TreeItem>): TreeItem => {
```
bggolding, line 29: We can do more of this in line with no real loss of readability:
```typescript
export const newTreeItem = (initFields: Pick<TreeItem, "type"> & Partial<TreeItem>): TreeItem => {
    return {
        id: initFields.id ?? Date.now(),
        content: "",
        instanceID: initFields.instanceID ?? `${id}-${0}`;
        ...initFields
    }
};
```

bggolding: Which could then be:
```typescript
export const newTreeItem = (initFields: Pick<TreeItem, "type"> & Partial<TreeItem>): TreeItem => ({
    id: initFields.id ?? Date.now(),
    content: "",
    instanceID: initFields.instanceID ?? `${id}-${0}`;
    ...initFields
});
```

src/components/context/FileProvider.tsx
```diff
@@ -26,12 +26,16 @@ export interface TreeNode {
...
+    const id = initFields.id ?? Date.now();
+    const instanceID = initFields.instanceID ?? `${id}-${0}`;
+    return{
+        id: id,
+        content: "",
+        instanceID:instanceID,  // give 0 when it is empty
```
bggolding, line 35: I don't understand what this comment means.

src/components/context/FileProvider.tsx
```diff
@@ -231,9 +235,9 @@ const FileProvider: React.FC<PropsWithChildren> = ({ children }) => {
     useEffect(() => {
         // Convert TreeNode[] to TreeItem[] for storage
         // Here we map TreeNode.goalId to TreeItem from state.goals
-        const treeItems = state.tree.map((treeNode:TreeNode) => {
+        const treeItems:TreeItem[] = state.tree.map((treeNode:TreeNode) => {
```
bggolding, line 238: Please put a space after a colon.

src/components/context/FileProvider.tsx
```diff
@@ -231,9 +235,9 @@ const FileProvider: React.FC<PropsWithChildren> = ({ children }) => {
...
             const goal = state.goals[treeNode.goalId];
-            if (!goal) return null;
+            if (!goal) return;
             return {
```
bggolding, line 241: There is a function `newTreeItem` which you could use to create the TreeItems so that they're of the correct type which would remove the `... as TreeItem[]`.

Are there really null elements in that list that need to be stripped out with the `.filter(Boolean)`?

src/components/context/treeDataSlice.ts
```diff
@@ -23,23 +22,13 @@ export const newTreeNode = ({goalId, instanceID,children = []}: {
 
 export const createTreeFromTreeData = (
   treeData: TreeItem[],
-  existingOcurrance: Record<TreeItem["id"], number> = {}
```
bggolding: Spelling error: should be `existingOccurrence`.

src/components/context/treeDataSlice.ts
```diff
@@ -23,23 +22,13 @@ export const newTreeNode = ({goalId, instanceID,children = []}: {
...
 ): TreeNode[] => {
-  
-  const existingDictionary = (goalId: TreeItem["id"]): number => {
-    if (existingOcurrance[goalId]===undefined) {
```
bggolding: Can `goalId` really be undefined? According to the type signature of this function, it cannot be.

src/components/context/treeDataSlice.ts
```diff
@@ -146,11 +135,19 @@ const generateInstanceID = (treeIds:Record<TreeItem["id"], TreeItem["instanceID"
   if(treeIds[goalId]==undefined){
     treeIds[goalId]=[]
   }
+
   // give it new instance id 
-  const instanceId: TreeItem["instanceID"] = (treeIds[goalId].length > 0)
-    ? Math.max(...treeIds[goalId]) + 1
-    : 1;
-  return instanceId
+  const maxSuffix:number = treeIds[goalId].length > 0
```
bggolding, line 140: You should make this into a function so that the logic of how the `instanceID` symbols are created can be localised to a single place, tested properly and reused.

Put a space after the colon.

Do not check in `console.log` statements.

src/data/initialTabs.ts
```diff
@@ -12,62 +12,66 @@ export interface InitialTab {
...
 }
 
-// TODO: this is a duplicate copy of this function to avoid a circular import -- fix this!!
-const newTreeItem = (initFields: Pick<TreeItem, "type"> & Partial<TreeItem>): TreeItem => ({
-    id: initFields.id ?? Date.now(),
-    content: "",
-    instanceID:initFields.instanceID??1,
-    ...initFields
-});
+//TODO: this is a duplicate copy of this function to avoid a circular import -- fix this!!
```
bggolding, line 15: Duplicating the fix in two places is bad. I know I did a bad thing leaving a duplicate copy of this function but this should have been an opportunity to fix this, not continue my bad example. 😞

src/data/initialTabs.ts
```diff
@@ -12,62 +12,66 @@ export interface InitialTab {
...
+    }
+};
 
 // Predefined constant cluster to use for the example graph
 export const defaultTreeData: TreeItem[] = [{
     id: 1,
     content: "Do",
     type: "Do",
-    instanceID:1,
+    instanceID:"1-1",
```
bggolding, line 32: Seeing this makes me think that the `newTreeItem` function could easily manufacture an `instanceID` attribute if one is not included in the arguments. This would simplify this code and hide its generation. Maybe ...

PR #149, opened 2025-09-28
96 goals should be draggable if placed in different hierarchy branches show duplicate warning only if in same branch

ToriLi1210:
Differentiate original goals from tree references and support multiple references

Description:

This PR refactors how goals are handled between the goal list and the tree to support multiple references of the same goal.

Problem:

Previously, each tree item used only id as its primary key.

This made it impossible to distinguish between the original goal in the goal list and its multiple references in the tree.

Deleting a goal or reference had inconsistent effects: deleting a reference sometimes affected other references.

Solution:

Differentiate the original goal (in the goal list) and tree references:

The goal in the list is the original.

The goal in the tree is a reference to the original.

Deleting a goal deletes all its references.

Deleting a reference only affects that reference (editing exceptions apply).

Introduce instanceID to each TreeItem to differentiate multiple references of the same goal.

Reformatted the treeIds structure: previously it was a list of TreeItem["id"]. Now it is a record mapping each TreeItem["id"] to an array of instanceIDs:

Record<TreeItem["id"], TreeItem["instanceID"][]>

This allows multiple references of the same goal to coexist in the tree without key conflicts.

Additional Notes:

Original tree item behavior is preserved for single-reference goals.

Dragging, editing, and deleting functionality respects the new reference system.

Prevents duplicate cloning when multiple references of a goal exist.

Changes:

Added instanceID to differentiate references.

Refactored treeIds to record multiple instanceIDs per goal.

Updated drag-and-drop logic and tree rendering to use instanceID as unique key

### DIFF
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -32,7 +32,7 @@ import {VERTEX_FONT} from "../utils/GraphConstants.tsx"
 import {removeGoalIdFromTree} from "../context/treeDataSlice.ts";
 import ConfirmModal from "../ConfirmModal.tsx";
 
-import {parseFunsGoalRefId} from "../utils/GraphUtils";
+import {parseFuncGoalRefId} from "../utils/GraphUtils";
 
 // ---------------------------------------------------------------------------
 
@@ -105,7 +105,7 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
         toRemove.forEach(cell => {
             const id = cell.getId();
             try {
-                const {goalId, instanceId} = parseFunsGoalRefId(id!); // parse throws on invalid
+                const {goalId, instanceId} = parseFuncGoalRefId(id!); // parse throws on invalid
                 parsedById.set(id!, {goalId, instanceId});
             } catch (err) {
                 invalids.push({
--- a/src/components/utils/GraphUtils.tsx
+++ b/src/components/utils/GraphUtils.tsx
@@ -22,7 +22,7 @@ export function formatFunGoalRefId(goal: ClusterGoal): string {
 }
 
 // Convert the cell id in MaxGraph 'Functional-8-1'
-export function parseFunsGoalRefId(idStr: string) {
+export function parseFuncGoalRefId(idStr: string) {
     if (!idStr) throw new Error("Cell ID is missing.");
 
     const parts = idStr.split("-");


src/components/Graphs/GraphHelpers.tsx
```diff
@@ -214,7 +214,6 @@ export const renderFunction = (
         height,
         style
     );
-    console.log("goalId:", goal.GoalID, " nodeId:", node.getId(), " value:", node.value);
```
bggolding, line 217: Thank you for removing this.

src/components/SectionPanel.tsx
```diff
@@ -57,7 +57,7 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
 
   const [draggedItem, setDraggedItem] = useState<TreeItem | null>(null);
   // Simply store ids of all items in the tree for fast check instead of recursive search
-  const {dispatch, treeIds} = useFileContext();
+  const {dispatch, treeIds,tree} = useFileContext();
```
bggolding, line 60: Space after comma please.

bggolding (changes requested):

src/data/initialTabs.ts
```diff
@@ -12,7 +12,6 @@ export interface InitialTab {
     rows: TreeItem[]
 }
 
-// //TODO: this is a duplicate copy of this function to avoid a circular import -- fix this!!
```
bggolding: Could you delete the copy of the function that's there, not just the comment.

src/components/context/treeDataSlice.ts
```diff
@@ -194,6 +189,11 @@ export const treeDataSlice = createSlice({
         },
         addGoalToTree: (state, action: PayloadAction<TreeItem>) => {
             // the instance id is given only when add to the tree
+
+            // if this is the first time the goal is moved into the hierarchy ✅
+            if(state.treeIds[action.payload.id]== undefined){
```
bggolding, line 194: This might read better as:
```typescript
if (!state.treeIds[action.payload.id]) {
    [...]
```

src/components/context/treeDataSlice.ts
```diff
@@ -194,6 +189,11 @@ export const treeDataSlice = createSlice({
...
+              state.treeIds[action.payload.id]=[]
+            }
```
bggolding, line 196: Actually, I think I'd rewrite this to be:
```typescript
const instanceID = generateInstanceID(state.treeIds, action.payload.id)
state.tree.push(newTreeNode({
    goalId: action.payload.id,
    instanceID: generateInstanceID(state.treeIds, action.payload.id)
}));
if (state.treeIds[action.payload.id]) {
     state.treeIds[action.payload.id].push(instanceID);
} else {
     state.treeIds[action.payload.id] = [instanceID];
}
```

bggolding: I would much rather see `instanceId` than `instanceID`.

src/components/context/treeDataSlice.ts
```diff
@@ -129,7 +129,7 @@ const removeAllReferenceFromHierarchy = (
     }));
 };
 
-const generateMaxSuffix  = (treeIds:Record<TreeItem["id"], TreeItem["instanceID"][]>,goalId:TreeItem["id"]):number=>{
+const generateMaxSuffix  = (treeIds: Record<TreeItem["id"], TreeItem["instanceID"][]>,goalId: TreeItem["id"]): number=>{
```
bggolding, line 132: Much better but I think you still need a space after the comma.

src/components/context/FileProvider.tsx
```diff
@@ -235,7 +235,7 @@ const FileProvider: React.FC<PropsWithChildren> = ({ children }) => {
     useEffect(() => {
         // Convert TreeNode[] to TreeItem[] for storage
         // Here we map TreeNode.goalId to TreeItem from state.goals
-        const treeItems: TreeItem[] =  createTreeDataFromTreeNode(state.goals,state.tree)
+        const treeItems = createTreeDataFromTreeNode(state.goals,state.tree)
```
bggolding, line 238: Space after comma please.

src/components/context/FileProvider.tsx
```diff
@@ -32,7 +32,7 @@ export const newTreeItem = (initFields: Pick<TreeItem, "type"> & Partial<TreeIte
     return{
         id: id,
         content: "",
-        instanceID:instanceID, 
+        instanceID: instanceID,
```
bggolding, line 35: (You don't actually need to repeat `instanceID` here, when you're initialising an object javascript will let you simply name it and it will use the value in the current context for the symbol's value. Like this:
```typescript
  return {
        id,
        content: "",
        instanceID, 
        ...initFields
    };
```
It's a very handy feature.)

src/components/SectionPanel.tsx
```diff
@@ -155,8 +155,7 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
 
       if (draggedItem && draggedItem.content) {
         // the first hierachy does not contain the dragged item
-          if (!tree.map((index)=>(index.goalId)).includes(draggedItem.id)) {
-            console.log("dragged item: ",draggedItem)
+          if (!tree.map((index) => (index.goalId)).includes(draggedItem.id)) {
```
bggolding, line 158: Please remove parens around `(index.goalId)`.

bggolding (changes requested):

PR #150, opened 2025-09-30
fix(graph): #82 bugitem created in the canvas is not saved or updated to the goal list

### LINKED ISSUE #82: Bug:Item created in the canvas is not saved or updated to the Goal List

KerrYuan:
### What does this PR do?
- Adjust cell ID assignment according to the new Functional/Nonfunctional + numeric ID rule.
- Dispatch newly created cells to both tabData and treeData to keep state in sync.

### Changes
- Update GeometryChange listener to handle new cell ID format.
- Ensure new cells are added to Redux store for both tabs and tree.

### Remaining problem
1. Dragged positions are not preserved after tree update.
  Current workflow: drag new node → tree update → re-render graph.
  Each tree update triggers a full graph re-render, which overwrites the node position set by drag-and-drop.
  
2. Edge(Line) cannot be added in canvas

### Related issue
Closes #82

### DIFF
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -27,8 +27,8 @@ import {useGraph} from "../context/GraphContext";
 import {Cluster} from "../types.ts";
 import GraphSidebar from "./GraphSidebar";
 import WarningMessage from "./WarningMessage";
-
-import {VERTEX_FONT, getSymbolConfigByShape,} from "../utils/GraphConstants.tsx"
+import {VERTEX_FONT} from "../utils/GraphConstants.tsx"
+import {getSymbolConfigByShape} from "../utils/GraphUtils";
 import {removeGoalIdFromTree, addGoalToTree, addGoal} from "../context/treeDataSlice.ts";
 import ConfirmModal from "../ConfirmModal.tsx";
 
@@ -227,11 +227,9 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
         evt.consume();
         try {
           const changes = evt.getProperty("edit").changes;
-      
           for (let i = 0; i < changes.length; i++) {
             const change = changes[i];
             if (change.constructor.name == "GeometryChange") {
-
               const cell: Cell = changes[i].cell;
               let cellId = cell.getId();
               let numericId: number | undefined;
@@ -239,7 +237,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
               if (cellId && /^(Functional|Nonfunctional)-(-?\d+)$/.test(cellId)) {
                 const match = cellId.match(/^(Functional|Nonfunctional)-(-?\d+)$/);
                 if (match) {
-                  numericId = Number(match[2]); // 这里 match[2] 就是数字部分，带负号
+                  numericId = Number(match[2]); 
                 }
               }
               else if (cellId && /^-?\d+$/.test(cellId)){
@@ -262,17 +260,13 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
                   id: numericId,
                   content: "",
                   type: cellLabel as Label,
-
-                  //children?: TreeItem[];
                 };
                 dispatch(addGoal(newTreeItem));
                 dispatch(addGoalToTree(newTreeItem));
-
               }
 
               if (cellID != null && cellHistory[cellID] == undefined) {
                 cellHistory[cellID] = [newWidth, newHeight];
-                
               } else {
                 if (cellID != null) {
                   const oldWidth = cellHistory[cellID][0];
@@ -303,8 +297,6 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
                 cellHistory[cellID] = [newWidth, newHeight];
               }
             }
-           
-
             else if (change.constructor.name == "ValueChange") {
               const cell: Cell = change.cell;
               // goal id
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -227,21 +227,39 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
         evt.consume();
         try {
           const changes = evt.getProperty("edit").changes;
+      
           for (let i = 0; i < changes.length; i++) {
             const change = changes[i];
             if (change.constructor.name == "GeometryChange") {
+
               const cell: Cell = changes[i].cell;
-              const numericId = Number(cell.getId());
+              let cellId = cell.getId();
+              console.log("！！！GeometryChange: cellId ",cellId)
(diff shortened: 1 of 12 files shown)


src/components/Graphs/GraphWorker.tsx
```diff
@@ -229,14 +232,32 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
             if (change.constructor.name == "GeometryChange") {
               const cell: Cell = changes[i].cell;
               const cellID = cell.getId();
-              console.log("change, ",cell)
+              console.log("！！！change, ",cell)
```
bggolding: Please do not commit `console.log` diagnostics.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -229,14 +232,32 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
...
               const oldStyle = cell.getStyle();
               const newWidth = cell.getGeometry()?.height;
               const newHeight = cell.getGeometry()?.width;
+            
+              const numericId = Number(cellID?.split("-").pop());
+              if (!treeIdsRef.current.includes(numericId)) {
+                console.log("！！！New goal added with id: ", numericId, "treeIds: ",treeIds.slice());
+                const newTreeItem: TreeItem = {
```
bggolding, line 259: You should use the `newTreeItem` function so that `TreeItem`s are created consistently and have the correct type.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -229,14 +232,32 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
             if (change.constructor.name == "GeometryChange") {
               const cell: Cell = changes[i].cell;
               const cellID = cell.getId();
-              console.log("change, ",cell)
+              console.log("！！！change, ",cell)
               const oldStyle = cell.getStyle();
               const newWidth = cell.getGeometry()?.height;
               const newHeight = cell.getGeometry()?.width;
+            
+              const numericId = Number(cellID?.split("-").pop());
```
bggolding: You should make a separate function to pull these id's apart so that the code is consolidated in one place where it can be tested and reused.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -285,7 +309,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
                 });
               } else {
                 for (let i = 0; i < lengthUpdated; i++) {
```
bggolding, line 318: You can get rid of `lengthUpdated` by changing this to:
```typescript
for (const id of cellIds.map((cellID) => cellID.split("-").pop()) {
    ...
}
```

bggolding (changes requested):

src/components/Graphs/GraphWorker.tsx
```diff
@@ -56,7 +56,10 @@ interface GlobObject {
 
 const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({ showGraphSection = false }) => {
   const divGraph = useRef<HTMLDivElement>(null);
-  const {cluster, dispatch} = useFileContext();
+  const {cluster, dispatch, treeIds} = useFileContext();
+  const treeIdsRef = useRef(treeIds);
+  useEffect(() => { treeIdsRef.current = treeIds }, [treeIds])
```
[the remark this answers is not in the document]
KerrYuan: I tried without the `useEffect`, but then the `ref` only captured the initial `treeIds` and later operations didn’t work correctly. The `useEffect` keeps it in sync with the latest value.

KerrYuan:
> "Check that an added be goal, feel goal and who are associated with a node in the do goal tree" from Leon

Currently, all the unfunctional goals created from canvas will automatically be related to the root goal.
The functional goal will be another root.
image

Adjusting/recognizing hierarchy via canvas should probably be a new user story.

src/components/utils/GraphUtils.tsx
```diff
@@ -6,11 +7,70 @@ export function getSymbolKeyByType(type: string): SymbolKey | undefined {
...
+  if (!cellId) return [];
+
+  // Remove prefix and split by comma
+  const match = cellId.match(/^(Functional|Nonfunctional)-(.+)$/);
+  if (!match) return [];
+
+  return match[2]
+    .split(",")
+    .map(s => Number(s.trim()))
+    .filter(n => !isNaN(n));
```
bggolding, line 69: If we've passed the `ensureCellIdFormat` test, how can there be a NaN?

src/components/Graphs/GraphWorker.tsx
```diff
@@ -297,30 +283,33 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
...
+            else if (change.constructor.name === "ValueChange") {
               const cell: Cell = change.cell;
-              // goal id
-              const cellIDs = cell.getId()?.split(",") ?? [];
-              // goal value
-              const newContent = change.value.split(",");
 
-              const lengthUpdated = cellIDs.length
+              // Ensure cell ID(s) are in proper format
+              ensureCellIdFormat(cell);
```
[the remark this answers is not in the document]
KerrYuan: It will throw an exception within the `ensureCellIdFormat`  function: ` else throw new Error(`Invalid cell ID format: ${part}`);`
I'm unsure if there is a better name for this function...

```JAVA
export function ensureCellIdFormat(cell: Cell): Cell {
  const cellId = cell.getId();

  const goalTypeRaw = getSymbolConfigByShape(String(cell.style.shape))?.type;
  const functionalType = goalTypeRaw === "Functional" ? "Functional" : "Nonfunctional";

  if (!cellId) {
    // Generate a new ID if none exists
    const newId = `${functionalType}-${Date.now()}`;
    cell.setId(newId);
    return cell;
  }

  // Split by comma and normalize each part
  const parts = cellId.split(",").map(p => p.trim()).filter(p => p.length > 0);
  const numericIds: number[] = parts.map(part => {
    const match = part.match(/^(Functional|Nonfunctional)-?(-?\d+)$/);
    if (match) return Number(match[2]);
    else if (/^-?\d+$/.test(part)) return Number(part);
    else throw new Error(`Invalid cell ID format: ${part}`);
  });

  // Join numeric IDs with comma, only first one has prefix
  const newId = `${functionalType}-${numericIds.join(",")}`;
  cell.setId(newId);
  return cell;
}`

File: src/components/utils/GraphUtils.tsx [hunk 8]
```diff
@@ -39,16 +39,12 @@ export const returnFocusToGraph = () => {
...
-    } else {
-        throw new Error("Functional goal expected");
-    }
+export const generateFunctionalCellId = (goal: number): string => {
+    return `Functional-${goal}`;
 };
 
 export const generateNonFunctionalCellId = (ids: number[]): string => {
-    return "Nonfunctional-" + ids.join(",");
+    return `Nonfunctional-${ids.join(",")}`;
```
bggolding: I like this much better.

PR #154, opened 2025-09-30
chore(error message): make the input error message clearer

### LINKED ISSUE #135: Rewrite message when feeling goals
Instead of "split by commas", it would be better if it said "separated by commas".

lvqq2000:
image

closes #135

### DIFF
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -20,14 +20,14 @@ import ErrorModal, {ErrorModalProps} from "../ErrorModal.tsx";
 import {associateNonFunctions, isGoalNameEmpty, layoutFunctions, renderGoals} from './GraphHelpers';
 import {registerCustomShapes} from "./GraphShapes";
 import "./GraphWorker.css";
-import {useFileContext} from "../context/FileProvider.tsx";
+import {Label, newTreeItem, useFileContext} from "../context/FileProvider.tsx";
 import {useGraph} from "../context/GraphContext";
 import {Cluster} from "../types.ts";
 import GraphSidebar from "./GraphSidebar";
 import WarningMessage from "./WarningMessage";
-
 import {VERTEX_FONT} from "../utils/GraphConstants.tsx"
-import {removeGoalIdFromTree} from "../context/treeDataSlice.ts";
+import {getCellNumericIds} from "../utils/GraphUtils";
+import {removeGoalIdFromTree, updateTextForInstanceId} from "../context/treeDataSlice.ts";
 import ConfirmModal from "../ConfirmModal.tsx";
 import {returnFocusToGraph} from "../utils/GraphUtils.tsx";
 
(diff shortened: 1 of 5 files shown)


src/components/Graphs/GraphWorker.tsx
```diff
@@ -280,7 +280,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
                 setErrorModal({
                   show: true,
                   title: "Input Error",
-                  message: `Please provide ${lengthUpdated} items split by comma`,
+                  message: `Please provide ${lengthUpdated} items separated by commas`,
```
bggolding, line 283: If we're going to make this change, we should just extend it so that the grammar is correct when there's only one item:
```typescript
`Please provide ${lengthUpdated} ${(lengthUpdated === 1) ? "item" : "items"} separated by commas`
```

bggolding (changes requested):

src/components/Graphs/GraphWorker.tsx
bggolding: As an aside, when you have a count for a list, I like to to use the shorthand like `nUpdated` because it links back to the original value and is nice and short. `lengthUpdated` is okay but ...

src/components/Graphs/GraphWorker.tsx
```diff
@@ -272,24 +272,24 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
...
+              const nUpdated = cellIDs.length
               // Check if the number of items matches
-              if (lengthUpdated !== newContent.length) {
+              if (nUpdated !== newContent.length) {
                 graph.getDataModel().setValue(cell, change.previous);
                 setErrorModal({
                   show: true,
                   title: "Input Error",
-                  message: `Please provide ${lengthUpdated} items split by comma`,
+                  message: `Please provide ${nUpdated} ${(nUpdated === 1) ? "item" : "items"} separated by commas`,
```
bggolding: That looks good.

src/components/Graphs/GraphWorker.tsx
bggolding: It's bad creating the payload for `dispatch` by hand: you lose all the type checking that's available; instead, use one of the actions that's defined in `treeDataSlice`. This should be:
```typescript
dispatch(updateTextForGoalId({id, text}));
```
(it doesn't look like you made the change but this is something that needs to happen everywhere that `dispatch` is called)

src/components/Graphs/GraphWorker.tsx
```diff
@@ -272,24 +272,24 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
...
                   onHide: () => setErrorModal(prev => ({ ...prev, show: false }))
                 });
               } else {
-                for (let i = 0; i < lengthUpdated; i++) {
-                  const id = Number(cellID[i]);
-                  const text = newContent[i].trim("");
+                for (const [index, cellID] of cellIDs.map(Number).entries()) {
```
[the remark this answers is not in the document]
bggolding: The other question I have is whether `newContent` is a good name for that variable. It's not clear what it holds.

lvqq2000:
@bggolding I have created another pr to fix this issue. May you review it?

PR #163, opened 2025-10-29
fix: correct the var name

lvqq2000:
@bggolding Ben I have changed the name 'newContent' to 'newGoalValues' in the last pr but I missed one instance. This PR updates the one I missed to use the new variable name 'newGoalValues'

### DIFF
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -285,7 +285,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
                   });
               } else {
                   numericCellIds.forEach((instanceId, i) => {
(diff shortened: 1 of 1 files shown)


src/components/Graphs/GraphWorker.tsx
```diff
@@ -290,7 +290,7 @@ const deleteItemFromGraph = (graph:Graph, removeChildrenFlag: boolean) => {
               } else {
                   numericCellIds.forEach((instanceId, index) => {
 
-                    const text = newContent[index];
+                    const text = newGoalValues[index];
```
bggolding: In all honesty, we con probably get rid of the `text`variable and just make the body of `forEach` into:
```typescript
 dispatch(updateTextForInstanceId({instanceId, text: newGoalValues[index]}));
```

I'd probably even take that one small step further and rename `index` to be `i`.

PR #174, opened 2025-11-05
165 change the wording of the delete warning window

ToriLi1210:


src/components/Graphs/GraphWorker.tsx
```diff
@@ -623,7 +623,7 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
             <ConfirmModal
                 show={showDeleteWarning}
                 title="Delete goal with children"
-                message="The selected goal has children. Confirm you want to delete this goal"
+                message="The selected goal has associate goal(s). Confirm you want to delete this goal"
```
bggolding: Is there a way that we can count the number of associated goals so that we can produce an error message that has correct English?

bggolding (changes requested):

bggolding (changes requested):
I think rather than creating another `useState` value which you are using to communicate between the `supportFunctions` method and the `ConfirmModal`, it would be better to create a small method called `childrenOfSelectedNode`.

```typescript
const childrenOfSelectedNode = (graph: Graph, selectedCell: Cell): Cell[] => {
    const outgoingEdges = graph.getOutgoingEdges(selectedCell);

    return outgoingEdges.filter((edge) => edge?.target !== selectedCell);
};
```

In `supportFunctions`:
```typescript
    const hasChildren = childrenOfSelectedNode(graph, selectedCells[0]).length > 0;
```

When you're getting the message together for the modal:
```typescript
const nAssociatedGoals = childrenOfSelectedNode(graph, selectedCells[0]).length;
```

There's little bit of recalculation here but it should be very small.

There are way too many `useEffect` calls in this component. I suspect a whole lot of them aren't necessary.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -625,8 +626,9 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
                 title="Delete goal with children"
                 message={
                     <>
-                        The selected goal has associated goal(s). <br />
-                        Confirm you want to delete this goal.
+                        The selected goal has {nAssociatedGoal} associated goal{nAssociatedGoal === 1 ? '' : 's'}. 
+                        <br/>
```
bggolding: It's a bit easier to read if you include whole words:
```typescript
The selected goal has {nAssociatedGoal} associated {(nAssociatedGoal === 1) ? 'goal' : 'goals'}
```

src/components/Graphs/GraphWorker.tsx
```diff
@@ -393,12 +396,11 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
...
 
-                const outgoingEdges = graph.getOutgoingEdges(selectedCells[0], null);
-                const hasChildren = outgoingEdges.some(edge => edge.target && edge.target !== selectedCells[0]);
-
+                const outgoingEdges = childrenOfSelectedNode(graph,selectedCells[0]);
+                const nAssociatedGoal = outgoingEdges.length;
                 // setRemoveChildren(hasChildren);
-                if (hasChildren) {
-                    setNAssociateGoal(outgoingEdges.length);
+                if (nAssociatedGoal>0) {
```
bggolding: Please put spaces around binary operators, ie, `nAssociatedGoal > 0`. Get rid of the empty line.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -379,6 +378,10 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
      * (1) Delete
      * (2) Undo
      */
+    const childrenOfSelectedNode = (graph: Graph, selectedCell: Cell): Cell[] => {
```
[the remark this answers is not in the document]
bggolding: Just reading the function declaration, it's jarring to me to see `childrenOfSelectedNode` and then to have the argument to the function be called `selectedCell`. Specifically, `node` != `cell`.

I think we should change one or the other to be `childrenOfSelectedCell` or the argument to `selectedNode`. I would probably lean towards the former but I think you could make a case either way.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -393,12 +396,11 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
 
 
 
-                const outgoingEdges = graph.getOutgoingEdges(selectedCells[0], null);
-                const hasChildren = outgoingEdges.some(edge => edge.target && edge.target !== selectedCells[0]);
-
+                const outgoingEdges = childrenOfSelectedNode(graph,selectedCells[0]);
+                const nAssociatedGoal = outgoingEdges.length;
```
bggolding: I was going to suggest removing the `outgoingEdges` variable because it's not strictly necessary but it's actually a nice way of documenting what that value represents. I think leave it.

bggolding (changes requested):

src/components/Graphs/GraphWorker.tsx
```diff
@@ -392,15 +394,12 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
                 const selectedCells = graph.getSelectionCells();
                 if (!selectedCells || selectedCells.length === 0) return;
 
-                deletingItemRef.current = selectedCells;
-
-
+                setDeletingCells(selectedCells);
 
-                const outgoingEdges = childrenOfSelectedNode(graph,selectedCells[0]);
+                const outgoingEdges = childrenOfSelectedCell(graph,selectedCells[0]);
```
bggolding: Space after comma please.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -74,11 +74,17 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
 
     const [showDeleteWarning, setShowDeleteWarning] = useState(false);
     const [removeChildren, setRemoveChildren] = useState(false);
-    const deletingItemRef = useRef<Cell[] | null>(null);
+    const [deletingCells, setDeletingCells] = useState<Cell[] | null>(null);
+
```
bggolding: Single blank line here please.

PR #175, opened 2025-11-05
fix: prevent non-functional goals from having child goals

### LINKED ISSUE #166: Prevent non-functional goals from having child goals
Currently, it’s possible to drag a goal to become a child of a non-functional goal in the hierarchy view. This should not be allowed, since non-functional goals cannot have child goals.

lvqq2000:
The user won't be able to drag goals to the right sides of the non-functional goals now

closes #166

### DIFF
--- a/src/components/context/FileProvider.tsx
+++ b/src/components/context/FileProvider.tsx
@@ -89,7 +89,7 @@ export type NonFunctionalGoalType = (typeof NON_FUNCTIONAL_GOAL_TYPES)[number];
 export const isNonFunctionalGoal = (
   label: Label | undefined
 ): label is NonFunctionalGoalType =>
-  !!label && NON_FUNCTIONAL_GOAL_TYPES.includes(label as NonFunctionalGoalType);
+  NON_FUNCTIONAL_GOAL_TYPES.includes(label as NonFunctionalGoalType);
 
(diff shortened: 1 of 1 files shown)


bggolding (changes requested):

src/components/context/FileProvider.tsx
```diff
@@ -86,8 +86,10 @@ export const NON_FUNCTIONAL_GOAL_TYPES = ["Be", "Feel", "Concern", "Who"] as con
 
 export type NonFunctionalGoalType = (typeof NON_FUNCTIONAL_GOAL_TYPES)[number];
 
-export const isNonFunctional = (label: Label): label is NonFunctionalGoalType =>
-  NON_FUNCTIONAL_GOAL_TYPES.includes(label as NonFunctionalGoalType);
+export const isNonFunctionalGoal = (
+  label: Label | undefined
+): label is NonFunctionalGoalType =>
+  !!label && NON_FUNCTIONAL_GOAL_TYPES.includes(label as NonFunctionalGoalType);
```
bggolding: Do you need the `!!label` in this test? `NON_FUNCTIONAL_GOAL_TYPES` will never include `undefined`

bggolding (approved):

PR #176, opened 2025-11-05
feat(shape): update the shape when there are multiple stakeholders

### LINKED ISSUE #160: When there are multiple stakeholders in the same layer, the stakeholder image in the model should be different from when there is only one stakeholder
When there are multiple stakeholders in the same layer, Image should change to something similar to Image to indicate there are multiple stakeholders

lvqq2000:
image

closes #160

### DIFF
--- a/src/components/Graphs/GraphHelpers.tsx
+++ b/src/components/Graphs/GraphHelpers.tsx
@@ -349,6 +349,23 @@ const adjustHorizontalPositions = (node: Cell, source: Cell, graph: Graph) => {
     graph.getDataModel().setGeometry(node, nodeGeo);
 };
 
+// Get the config type based on the type and the descriptions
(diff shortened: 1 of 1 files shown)


src/components/Graphs/GraphHelpers.tsx
```diff
@@ -403,6 +411,11 @@ export const renderNonFunction = (
                 y = geo.y + OFFSET_Y;
                 delimiter = "\n";
                 break;
+            case "CROWD": // Same as stakehoder
```
bggolding: If it's the same as STAKEHOLDER, then just make both cases use the same code, ie:
```typescript
            case "STAKEHOLDER":
            case "CROWD":
                // Bottom Left
                x = geo.x - width - OFFSET_X;
                y = geo.y + OFFSET_Y;
                delimiter = "\n";
                break;
```

bggolding (approved):
Thanks, that looks good.

Issue #173, opened 2025-11-05
Colour information lost after editing canvas


ToriLi1210:
Any interaction that updates the canvas causes the element’s assigned colour to be lost.

PR #179, opened 2025-11-10
#173: Save color information after canvas update

lollyluan:
Added `color` field to `TreeNode` and `TreeItem` interfaces
- Allows each goal instance to maintain its own color state.

### DIFF
--- a/src/components/Graphs/SidebarComponents/ColorButtons.tsx
+++ b/src/components/Graphs/SidebarComponents/ColorButtons.tsx
@@ -6,7 +6,7 @@ import ButtonGroup from "react-bootstrap/ButtonGroup";
 import Button from "react-bootstrap/Button";
 import ColorPicker from "./ColorPicker.tsx";
 import {useFileContext} from "../../context/FileProvider.tsx";
-import {parseFuncGoalRefId} from "../../utils/GraphUtils.tsx";
+import {parseFuncGoalRefId, parseGoalRefId} from "../../utils/GraphUtils.tsx";
 import {updateColorForInstanceId} from "../../context/treeDataSlice.ts";
 
 type ColorButtonsProps = {
@@ -21,7 +21,7 @@ const ColorButtons = ({graph}: ColorButtonsProps) => {
         try {
             graph.getSelectionCells().forEach((cell) => {           
                 const id = cell.getId();
-                const instanceId = parseFuncGoalRefId(id!)?.[0].instanceId;
+                const instanceId = parseGoalRefId(id!)?.[0].instanceId;
                 if (instanceId) {
                     dispatch(updateColorForInstanceId({instanceId, color}));
                 }
--- a/src/components/utils/GraphUtils.tsx
+++ b/src/components/utils/GraphUtils.tsx
@@ -1,7 +1,7 @@
 import {ClusterGoal, GoalBase} from '../types';
 import {SYMBOL_CONFIGS, SymbolKey, SymbolConfig} from './GraphConstants';
 import {Graph, Cell} from '@maxgraph/core';
-import {TreeNode} from "../context/FileProvider.tsx";
+import {TreeNode} from "../../data/dataModels";
 
 // Finds the symbol key (e.g. 'STAKEHOLDER') based on the type
 export function getSymbolKeyByType(type: string): SymbolKey | undefined {
--- a/src/components/Graphs/GraphHelpers.tsx
+++ b/src/components/Graphs/GraphHelpers.tsx
@@ -15,7 +15,7 @@ import {
 } from "../utils/GraphConstants.tsx";
 
 import {getSymbolKeyByType, formatFunGoalRefId, generateCellId, getNonFunctionalGoalColor, makeLabelForGoalType} from "../utils/GraphUtils";
-import {TreeNode} from "../context/FileProvider.tsx";
+import {TreeNode} from "../../data/dataModels.ts";
 
 // ---------------------------------------------------------------------------
 // some image path
--- a/src/data/dataModels.ts
+++ b/src/data/dataModels.ts
@@ -7,12 +7,14 @@ export interface TreeItem {
   type: Label;
   instanceId: string;
   children?: TreeItem[];
+  color?: string;
 }
 
 export interface TreeNode {
   goalId: TreeItem["id"];
   instanceId: TreeItem["instanceId"];
   children?: TreeNode[];
+  color?: TreeItem["color"];
 }
 
 export interface TabContent {


src/components/Graphs/SidebarComponents/ColorButtons.tsx
```diff
@@ -5,20 +5,26 @@ import {ColorResult} from "react-color";
...
         try {
-            graph.getSelectionCells().forEach((cell) => {
-                const style = graph.getCellStyle(cell);
-                style.fillColor = color;
-                graph.getDataModel().setStyle(cell, style);
+            graph.getSelectionCells().forEach((cell) => {           
+                const id = cell.getId();
+                const instanceId = parseFuncGoalRefId(id!)?.[0]?.instanceId ?? null;
+                if (instanceId){
+                    dispatch(updateColorForInstanceId({instanceId, color: color}));
```
bggolding: This can just be:
```typescript
dispatch(updateColorForInstanceId({instanceId, color}));
```

src/components/context/treeDataSlice.ts
```diff
@@ -153,6 +154,18 @@ export const createInitialState = (tabData: InitialTab[] = initialTabs, treeData
     };
 };
 
+const updateTreeNodeColor = (nodes: TreeNode[], goalId: number, instanceId: string, color: string) => {
+    for (const node of nodes) {
```
bggolding: Is it worth thinking about using `forEach` here?

src/components/context/treeDataSlice.ts
```diff
@@ -155,15 +155,15 @@ export const createInitialState = (tabData: InitialTab[] = initialTabs, treeData
 };
 
 const updateTreeNodeColor = (nodes: TreeNode[], goalId: number, instanceId: string, color: string) => {
-    for (const node of nodes) {
+    nodes.forEach((node) => {
```
bggolding: That looks good.

src/components/Graphs/SidebarComponents/ColorButtons.tsx
```diff
@@ -21,10 +21,8 @@ const ColorButtons = ({graph}: ColorButtonsProps) => {
         try {
             graph.getSelectionCells().forEach((cell) => {           
                 const id = cell.getId();
-                const instanceId = parseFuncGoalRefId(id!)?.[0]?.instanceId ?? null;
-                if (instanceId){
-                    dispatch(updateColorForInstanceId({instanceId, color: color}));
-                }
+                const instanceId = parseFuncGoalRefId(id!)?.[0].instanceId;
+                if (instanceId) dispatch(updateColorForInstanceId({instanceId, color: color}));
```
bggolding: Javascript's rules about line continuations are a mess. I strongly recommend wrapping lines which are subordinate to control flow constructs (like `if`/`for`/`while`/...) in braces so there's no confusion.

Note I changed how `color` is included in the object as well.
```typescript
if (instanceId) {
    dispatch(updateColorForInstanceId({instanceId, color}));
}
```

src/components/utils/GraphUtils.tsx
```diff
@@ -195,7 +195,7 @@ export function getNonFunctionalGoalColor(
     return undefined;
 }
 
-function findGoalbyInstanceId(clusterGoals: ClusterGoal[], instanceId: string): GoalBase | null{
+function findGoalbyInstanceId(clusterGoals: ClusterGoal[], instanceId: string): GoalBase | null {
```
bggolding: We should make this an arrow function to for consistency.

src/components/utils/GraphUtils.tsx
```diff
@@ -195,7 +195,7 @@ export function getNonFunctionalGoalColor(
...
     for (const goal of clusterGoals) {
```
bggolding: forEach?

src/components/context/treeDataSlice.ts
```diff
@@ -253,7 +261,10 @@ export const treeDataSlice = createSlice({
         }>) => {
             const {instanceId, color} = action.payload;
             const goalId = parseInstanceId(instanceId).goalId;
-            updateTreeNodeColor(state.tree, goalId, instanceId, color);
+            const node = findTreeNodeByInstanceId(state.tree, instanceId);
+            if (node && node.goalId === goalId) {
```
bggolding: This is a question, not necessarily a request, but do you think that we should log a warning or something if there is no node with that `instanceId`?

src/components/context/treeDataSlice.ts
```diff
@@ -154,16 +154,24 @@ export const createInitialState = (tabData: InitialTab[] = initialTabs, treeData
...
 };
 
-const updateTreeNodeColor = (nodes: TreeNode[], goalId: number, instanceId: string, color: string) => {
+const findTreeNodeByInstanceId = (nodes: TreeNode[], instanceId: string): TreeNode | undefined => {
+    let result: TreeNode | undefined;
+
     nodes.forEach((node) => {
-        if (node.goalId === goalId && node.instanceId === instanceId) {
-            node.color = color;
+        if (result) {
```
bggolding: At least make this:
```typescript
return undefined;
```
A `return` statement with no value makes it look like there is a `return` with no value in a function which also has a `return` with a value.

src/components/context/treeDataSlice.ts
```diff
@@ -154,16 +154,24 @@ export const createInitialState = (tabData: InitialTab[] = initialTabs, treeData
...
             return;
         }
+        if (node.instanceId === instanceId) {
+            result = node;
+        }
         if (node.children) {
-            updateTreeNodeColor(node.children, goalId, instanceId, color);
+            const found = findTreeNodeByInstanceId(node.children, instanceId);
```
bggolding: You can just:
```typescript
return findTreeNodeByInstanceId(node.children, instanceId);
```

src/components/context/treeDataSlice.ts
```diff
@@ -154,16 +154,24 @@ export const createInitialState = (tabData: InitialTab[] = initialTabs, treeData
...
+    let result: TreeNode | undefined;
+
     nodes.forEach((node) => {
-        if (node.goalId === goalId && node.instanceId === instanceId) {
-            node.color = color;
+        if (result) {
             return;
         }
+        if (node.instanceId === instanceId) {
+            result = node;
```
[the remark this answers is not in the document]
bggolding: I started writing some unit tests to check whether this works but there's something funny going on and I couldn't get them to run.

lollyluan:
> 
> 
> 
> 
> > On 15 Nov 2025, at 23:50, Lolly Luan ***@***.***> wrote:
> 
> > 
> 
> > 
> 
> > @lollyluan commented on this pull request.
> 
> > 
> 
> > In src/components/context/treeDataSlice.ts <https://github.com/MotivationalModelling/mm-local-editor/pull/179#discussion_r2529868232>:
> 
> > 
> 
> > >              return;
> 
> >          }
> 
> > +        if (node.instanceId === instanceId) {
> 
> > +            result = node;
> 
> > I got the same issue, and just want to double check to run the test, we use npx vitest, is this right?
> 
> > image.png (view on web) <https://github.com/user-attachments/assets/593b0027-4449-4842-a3b9-5083f629bae8>
> 
> Yes, that’s the same error I’m seeing. I haven’t got time at the moment to dig into the issue (I’ve got a lot of marking to do at the moment).
> 
> 
> 
> I’m just running vitest from the command line (or using PyCharm’s test runner).
> 
> 
> 
> 	Ben.
> 
> 
> 
> 

I am trying to fix this problem. I've created a new testing file without importing `initialTabs`, however, it still comes out with this warning. I think it might be the issue with test environment.

bggolding:
> On 16 Nov 2025, at 14:06, Ben Golding ***@***.***> wrote:
> 
> Yes, that’s the same error I’m seeing. I haven’t got time at the moment to dig into the issue (I’ve got a lot of marking to do at the moment).
> 
> I’m just running vitest from the command line (or using PyCharm’s test runner).

I had a quick look and found a page which suggested that vitest doesn’t like arrow functions. I changed it to:

	export function newTreeItem(initFields: Pick<TreeItem, "type"> & Partial<TreeItem>): TreeItem {
and to my surprise it worked. This is bad and inexplicable behaviour.
	Ben.

src/components/context/treeDataSlice.ts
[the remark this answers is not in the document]
lollyluan: > I've had another look at the code and I think this should be close:
> 
> ```ts
> export const findTreeNodeByInstanceId = (nodes: TreeNode[], instanceId: TreeNode["instanceId"]): TreeNode | undefined => {
>     nodes.forEach((node) => {
>         if (node.instanceId === instanceId) {
>             return node;
>         }
>         const matchingNode = findTreeNodeByInstanceId(node.children ?? [], instanceId);
>         if (matchingNode) {
>             return matchingNode;
>         }
>     });
>     return undefined;
> };
> ```

Hi Ben,

I just added some unit test and test with the above function, but it always return the result with `undefined.`

image

bggolding:
I've made some changes and got the function and the tests working. Could you check it please?

Issue #167, opened 2025-11-05
Change Button Label When Color Palette Opens


lvqq2000:
When the color palette is opened, change the "Select Color" button to "Hide Color". When the user presses it, it will close the color palette window

PR #186, opened 2025-11-15
Make title of colour picker button change depending upon whether the …

bggolding:
…picker is displayed or not.

#167


bggolding:
Could you do the merge to develop please?

PR #188, opened 2025-11-17
180 unit tests have stopped working

ToriLi1210:


src/components/Tree.tsx
```diff
@@ -6,9 +6,9 @@ import FeelIcon from "/img/Heart.png";
 import ConcernIcon from "/img/Risk.png";
 import Nestable, {NestableProps} from "react-nestable";
 import {FaPlus, FaMinus} from "react-icons/fa";
-import {TreeItem} from "./context/FileProvider";
+import {TreeItem,Label} from "../data/dataModels.ts";
```
bggolding: This might be a stupid question but is `dataModels` a good name for this file? It's a very specific name for what I think of as a broad collection of things. Would `types` be better?

bggolding (approved):

PR #199, opened 2025-11-23
fix: fix the bug where sidebar with duplicate stakeholder symbol

lvqq2000:


bggolding (approved):
That looks good. I think omitting the "Crowd" symbol without changing the display logic is good.

PR #200, opened 2025-11-23
feat: add dot line to connect non-functional goal with functional goal

### LINKED ISSUE #190: Connecting non-functional and functional goals
Add a dotted line between non-functional goal and the associated functional goals.

This is an experimental feature to see how it will look.

lvqq2000:
image

closes #190

### DIFF
--- a/src/components/Graphs/GraphHelpers.tsx
+++ b/src/components/Graphs/GraphHelpers.tsx
@@ -22,8 +22,10 @@ import {TreeNode} from "../context/FileProvider.tsx";
 // const PATH_EDGE_HANDLER_ICON = "img/link.png";
 
 // default x,y coordinates of the root goal in the graph - (functional graph)
-const SYMBOL_X_COORD = 0;
-const SYMBOL_Y_COORD = 0;
+const DEFAULT_ROOT_GOAL_COORD = {
+    x: 0,
+    y: 0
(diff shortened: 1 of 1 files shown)


src/components/Graphs/GraphHelpers.tsx
```diff
@@ -431,6 +431,15 @@ export const renderNonFunction = (
     style.labelPosition = "center";
     style.spacingTop = 0;
 
+    // Clone edge style
+    const dotted: any = { ...graph.getStylesheet().getDefaultEdgeStyle() };
```
bggolding: A better way to do this is:
```typescript
const dotted = {
    ...graph.getStylesheet().getDefaultEdgeStyle(),
    dashed: 1,
    dashPattern: "1 3",
    rounded: 1
};
```

Personally, I think it would look better with dashes instead of dots.

bggolding (changes requested):
BTW, nice code.

bggolding (approved):
The changes you've made to the symbol names are *excellent*! The code is so much clearer, not just with the `.x` and `.y` but the other names as well.

Is it worth using `.x` and `.y` instead of `.horizontal` and `.vertical`? If you decided to do that, just reopen the branch, make the changes and assign a PR to me.

PR #227, opened 2025-12-19
191 remove treenode and use treegoal for later development

### LINKED ISSUE #191: Remove treeItem and use treeGoal for later development
deprecate treeItem

aoxiangx:
every commits have detail stated.
all tests passed, 58 in total (deprecated 1 and also added a new 1)


aoxiangx:
Updated and merged with the changes from latest develop branch, waiting for review.

src/components/types.ts
```diff
@@ -1,10 +1,16 @@
...
+// ============================================
+export type InstanceId = `${number}-${number}`
+export type Label = "Do" | "Be" | "Feel" | "Concern" | "Who";
+
 export type GoalType = "Functional" | "Quality" | "Stakeholder" | "Negative" | "Emotional"
 
 export interface GoalBase {
     GoalID: number
-    instanceId: TreeNode["instanceId"]
+    instanceId: InstanceId
```
bggolding: I would like to see this kept as `TreeGoal["instanceId"]`. That way,  if the definition of that type in`TreeGoal` changes, all the references to it will pick up that change automatically.

src/components/types.ts
```diff
@@ -117,41 +123,39 @@ export const GoalModelProjectSchema = z.object({
...
 
-// // Define the initial tabs with labels and corresponding icons
-export type InstanceId = `${number}-${number}`
-
-// Type of the tree item content
-export type TreeItem = {
+// ============================================
+// TreeGoal - the unified tree structure type
+// ============================================
+export type TreeGoal = {
```
bggolding, line 126: Does this comment really add anything?

src/components/types.ts
```diff
@@ -117,41 +123,39 @@ export const GoalModelProjectSchema = z.object({
...
-    children?: TreeNode[];
-    color?: TreeItem["color"];
-}
+// ============================================
+// DEPRECATED: TreeNode is now an alias for TreeGoal
+// Will be removed in future refactoring steps
+// ============================================
+/** @deprecated Use TreeGoal instead */
+export type TreeNode = TreeGoal;
+/** @deprecated Use newTreeGoal instead */
```
bggolding: Just delete it. Don't mark it as deprecated, just get it out of the code. The place for history in the project is in the change sets in the git repo, not in the code.

src/components/context/treeDataSlice.ts
```diff
@@ -1,52 +1,55 @@
...
         children = [],
+        content = "",
+        type,
+        color,
     }: {
-        goalId: TreeItem["id"];
-        instanceId?: TreeItem["instanceId"];
-        children?: TreeNode[];
+        id: TreeGoal["id"];
+        instanceId?: InstanceId;
```
bggolding: Use `TreeGoal["instanceId"]`.

src/components/context/treeDataSlice.ts
```diff
@@ -1,52 +1,55 @@
...
-        goalId: ti.id,
-        // if ti.instanceId exists, use it, else compute one
-        instanceId: ti.instanceId,
-        children: createTreeFromTreeData(ti.children ?? []),
-        color: ti.color
-    }));
-};
+
+/** @deprecated No longer needed - tree now stores TreeGoal directly */
+export const createTreeFromTreeData = (treeData: TreeGoal[]): TreeGoal[] => treeData;
```
bggolding: If it's no longer needed, delete it. Don't leave old cruft in the code -- it just makes more stuff people have to learn about to work in the codebase.

src/components/context/treeDataSlice.ts
```diff
@@ -130,80 +133,87 @@ const generateMaxSuffix = (treeIds: Record<TreeItem["id"], TreeItem["instanceId"
...
+export const createInitialState = (tabData: InitialTab[] = initialTabs, treeData: TreeGoal[] = []) => {
 
     const {goals, tabs} = createGoalsAndTabsFromTabContent(tabData);
 
     // console.log("createInitialState", tabContent, goals, tabs);
     return {
         tabs,
         goals,
-        tree: createTreeFromTreeData(treeData),
+        tree: treeData,  // No conversion needed - tree now stores TreeGoal directly
```
bggolding: The comment isn't necessary or helpful.

src/components/context/treeDataSlice.ts
```diff
@@ -230,7 +240,7 @@ export const treeDataSlice = createSlice({
             };
         },
         updateTextForInstanceId: (state, action: PayloadAction<{
-            instanceId: string,
+            instanceId: string,  // Accept string for compatibility with getCellNumericIds
```
bggolding: I would not do this. Either make the user cast the type to do this assignment or (better) write a little function that does a test that the InstanceId conforms to the correct pattern.

Something like:
```typescript
const checkInstanceId = (id: string): InstanceId => {
    // raise exception if id is in wrong format
    return id as InstanceId;
}
```

src/components/context/FileProvider.tsx
```diff
@@ -192,28 +173,25 @@ const FileProvider: React.FC<PropsWithChildren> = ({children}) => {
...
 
     return (
         <FileContext.Provider value={{
             ...state,
             dispatch,
-            treeData: computedTreeData,
+            treeData: state.tree,  // No conversion needed
             tabData: computedTabData,
-            cluster: convertTreeDataToClusters(state.goals, state.tree),
+            cluster: convertTreeDataToClusters(state.tree),  // Simplified - no goals param needed
```
bggolding: Not a useful comment. You are documenting how the code used to be, not how it is now.

src/components/context/FileProvider.test.tsx
```diff
@@ -118,51 +118,51 @@ describe('FileProvider', () => {
...
         ];
         const treeIds = createTreeIdsFromTreeData(treeData);
 
         expect(Object.keys(treeIds)).toEqual(["1","2"]);
     });
     it('should handle nested nodes', () => {
         const treeData = [
-                newTreeItem({type: "Do", id: 1, children: [
-                    newTreeItem({type: "Do", id: 2})
+                newTreeGoal({type: "Do", id: 1, children: [
```
bggolding: The indenting here looks not quite right.

src/components/context/treeDataSlice.test.ts
```diff
@@ -155,82 +154,71 @@ describe('treeDataSlice', () => {
...
+        const targetInstanceId: InstanceId = "3-1";
+        const targetId = 3;
+        expect(findTreeGoalByInstanceId(testTree, targetInstanceId)?.id).toBe(targetId);
     });
     it('should return undefined if node not found', () => {
-        expect(findTreeNodeByInstanceId(testTree, "not-exist")).toBeUndefined;
+        const nonExistentId: InstanceId = "999-999";
+        expect(findTreeGoalByInstanceId(testTree, nonExistentId)).toBeUndefined();
     });
-});
```
bggolding, line 236: Thank you for updating the unit tests.

bggolding (changes requested):
Looks pretty good but I'd like to see some changes.

aoxiangx:
The changes have been made based on your advice.
Looking forward to your review.

src/components/utils/GraphUtils.tsx
```diff
@@ -185,12 +185,24 @@ export function generateCellId<T extends keyof IdsForType>(type: T, ids: IdsForT
     }
 }
 
+export const validateInstanceId = (id: string): InstanceId => {
+    const bits = id.split("-").map(s => s.trim());
```
bggolding, line 189: This is all looking really good but if the instanceId has spaces in it then it's badly formatted and the code should complain.

Would it be better to use a regular expression to do this checking rather than pulling the instanceId to pieces? I seem to recall that there might already be a reg exp to do that somewhere.

bggolding (changes requested):
This is looking really quite good now. Thanks for all your work.

I am starting to wonder if all the pulling apart and validation of instanceId should be done by a reg exp rather than in code. It makes it easier to understand and validate.

src/components/utils/GraphUtils.tsx
bggolding: This should be a separate issue, not done as part of this one.

PR #228, opened 2025-12-22
114 update user tips

### LINKED ISSUE #114: Update the user tips
Right now the user tips for "Enter Goals / Arrange hierarchy" subpage and "Arrange hierarchy / Render Model" are out of date. They should be updated.

Image

Image

lvqq2000:
Updated the tips:
image
image

closes #114


src/components/InfoPopover/ClusterInfoPopover.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import {Popover} from "react-bootstrap";
+import {BsFillTrash3Fill, BsPlus, BsPencilSquare} from "react-icons/bs";
+
+
+export const ClusterInfoPopover = React.forwardRef<
```
bggolding: Can we call this `ClusterUsageInfo` please?

src/components/InfoPopover/ClusterInfoPopover.tsx
```diff
@@ -0,0 +1,48 @@
...
+  React.ComponentProps<typeof Popover>
+>((props, ref) => {
+  const { children, ...popoverProps } = props;
+
+  return (
+    <Popover
+      ref={ref}
+      {...popoverProps}
+    >
+    <Popover.Body>
```
bggolding: This is close but what I wanted to do was to make the component just have the text for the info on the page. I wanted the control over how that was displayed to be separate from the actual content, that is, the `<Popover>` stuff should be in the `ProgressBar` component.

I'd like it to look like:
```jsx
<Popover style={{ maxWidth: "max-content" }}>
    <Popover.Body>
         <ClusterUsageInfo/>
    </Popover.Body>
</Popover>
```

What I'm trying to achieve is to make the content separate from the layout. If the content changes, that shouldn't affect the layout; similarly, if the layout changes that shouldn't affect the content. If we made the usage info slide down from the top (say), we'd have to change the name of the component because it wouldn't be a Popover any more.

bggolding (changes requested):
I’m sorry to ask for another change but ...

bggolding (changes requested):

PR #229, opened 2025-12-28
#222 non functional overlay issue

### LINKED ISSUE #205: Non-functional label spacing needs to be closer
**_Non-functional label is placed too far away and requires an adjustable spacing variable_**

The non-functional label currently sits too far from the related element, creating unnecessary visual separation. 

**Expected Behaviour**

1. The non-functional label should be positioned closer to its associated element.

**Actual Behaviour**

The spacing is fixed and excessively large, with no way to adjust it.

Image

### LINKED ISSUE #222: Functional goal spacing causes overlap with non-functional goals
When there are multiple layers and many functional goals, the spacing between functional goals is small. As a result, when two functional goals each have their own non-functional goals, these non-functional goals can overlap.
Image

Here is the test model you can use to reproduce the issue:
[Model.json](https://github.com/user-attachments/files/24180478/Model.json)

Expected behavior:
Increase or dynamically adjust the spacing between functional goals to prevent non-functional goals from overlapping and improve readability.

Current behavior:
Functional goals are placed too close together, causing their associated non-functional goals to overlap and making the layout confusing.

lvqq2000:
image

closes #222 

closes #205

### DIFF
--- a/src/components/Graphs/GraphWorker.tsx
+++ b/src/components/Graphs/GraphWorker.tsx
@@ -535,8 +535,6 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
             stakeholdersGlob
         );
         graph.getDataModel().endUpdate();
-
-
     }, [graph, cluster]);
 
     // First useEffect to set up graph. Only run on mount.


bggolding (changes requested):
The code is looking better for this change but I've added a comment about naming variables.

I'm always looking for ways to remove `useEffect`. It is difficult to understand when the conditions will be met so that code will be executed. Instead I try to find what will trigger the change and then make that code execute there directly rather than as a side-effect of variables being updated. That's not always possible but it's what I aim for.

bggolding (approved):
Great.

PR #258, opened 2026-02-25
fix: the edge should toward the center of the shape to avoid disconne…

### LINKED ISSUE #204: Line connections incorrectly treated as a square
**_Line connections are misaligned due to the shape being treated as a square instead of a parallelogram_**

The current connection logic assumes the shape is a square, but the actual shape used in the interface is a parallelogram. As a result, the lines are incorrectly positioned and do not align with the intended geometry.

If possible, the contact point should be placed at the centre of the pattern, and the line should be routed from behind the pattern for a cleaner layout.

**Expected Behaviour**

1. Line connections should follow the geometry of a parallelogram.

2. The contact point should be positioned at the centre of the pattern.

3. The line should be routed from behind the pattern.

**Actual Behaviour**

The lines are aligned as if the shape were a square, leading to incorrect positioning.

Image

ToriLi1210:
…ction

### DIFF
--- a/src/components/Graphs/GraphHelpers.tsx
+++ b/src/components/Graphs/GraphHelpers.tsx
@@ -208,6 +208,8 @@ export const renderFunction = (
     // Make sure to specify what shape we're drawing
     style.shape = config.shape;
 
+    style.perimeter = "none"; // Disables the perimeter check so the line reaches the center
+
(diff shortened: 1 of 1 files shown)


bggolding (approved):
That looks good Tori -- nice work.

Issue #254, opened 2026-02-21
When adding a goal using drag and drop, the title isn't editable.


bggolding:
When I add a new goal using drag and drop from the symbol palette, it adds the symbol to the model but the name can't be edited in the arrange hierarchy; it can be edited using the goal list.

<img width="1402" height="640" alt="Image" src="https://github.com/user-attachments/assets/1886c39d-b082-455c-b25a-7f1986819cc3" />

RaghavDabra:
I investigated this issue and it seems to be related to how new goals are initialized when they are created through drag-and-drop.

In src/components/types.ts, the newTreeGoal() function initializes a goal with empty content:

```
export const newTreeGoal = (initFields: Pick<TreeGoal, "type"> & Partial<TreeGoal>): TreeGoal => {
    const id = initFields.id ?? Date.now();
    const instanceId = initFields.instanceId ?? `${id}-0`;
    return { id, content: "", instanceId, ...initFields };
};

```
Because the goal starts with content: "", it is treated as an empty goal. In GoalHint.tsx, editing appears to be restricted when the goal content is empty (isEmptyGoal() / canEditGoal()), which likely prevents editing the title in the Arrange Hierarchy view.

One possible fix would be to initialize dragged goals with a default title so they are editable immediately.

Example change:

```
export const newTreeGoal = (initFields: Pick<TreeGoal, "type"> & Partial<TreeGoal>): TreeGoal => {
    const id = initFields.id ?? Date.now();
    const instanceId = initFields.instanceId ?? `${id}-0`;
    return { id, content: "New Goal", instanceId, ...initFields };
};

```
Alternatively, the default content could be set in the drag-and-drop logic (e.g., in SidebarItems.tsx) when creating the goal.

RaghavDabra:
Hey Ben, so I dug into the drag-and-drop bug and it turned out to be three things that were basically looping (not sure if this is a right term) together rather than one. The first was that handleEdit had a guard blocking any goal with empty content 
->which made sense for phantom nodes but also caught brand-new goals that legitimately start empty. 

The second was that even without that guard, nothing was automatically putting the user into edit mode after the drop, so they'd still have to manually click the pencil icon. -> So with a useEffect that watches for newly added empty nodes and immediately focuses the input. 

The third was that if you dropped a goal and then pressed Escape without typing anything, the node just stayed in the tree permanently with no way to ever name it so I made cancel cleaning that up properly. 
[...]

PR #261, opened 2026-03-08
#254 - When adding a goal using drag and drop, the title isn't editable. 

RaghavDabra:
Fix would be to initialize dragged goals with a default title so they are editable immediately.

The goal starts with content: "", it is treated as an empty goal. In GoalHint.tsx, editing appears to be restricted when the goal content is empty (isEmptyGoal() / canEditGoal()), which likely prevents editing the title in the Arrange Hierarchy view.

In src/components/types.ts, the newTreeGoal() function initializes a goal with empty content:
```

export const newTreeGoal = (initFields: Pick<TreeGoal, "type"> & Partial<TreeGoal>): TreeGoal => {
    const id = initFields.id ?? Date.now();
    const instanceId = initFields.instanceId ?? `${id}-0`;
    return { id, content: "", instanceId, ...initFields };
};

```
Because the goal starts with content: "", it is treated as an empty goal. In GoalHint.tsx, editing appears to be restricted when the goal content is empty (isEmptyGoal() / canEditGoal()), which likely prevents editing the title in the Arrange Hierarchy view.

One possible fix would be to initialize dragged goals with a default title so they are editable immediately.

Change I made:
```

export const newTreeGoal = (initFields: Pick<TreeGoal, "type"> & Partial<TreeGoal>): TreeGoal => {
    const id = initFields.id ?? Date.now();
    const instanceId = initFields.instanceId ?? `${id}-0`;
    return { id, content: "New Goal", instanceId, ...initFields };
};
```

### DIFF
--- a/src/components/types.ts
+++ b/src/components/types.ts
@@ -135,7 +135,7 @@ export type TreeGoal = {
 export const newTreeGoal = (initFields: Pick<TreeGoal, "type"> & Partial<TreeGoal>): TreeGoal => {
     const id = initFields.id ?? Date.now();
     const instanceId = initFields.instanceId ?? `${id}-0`;
-    return {id, content: "", instanceId, ...initFields};
+    return { id, content: "New Goal", instanceId, ...initFields };
 };
 
 // Define the structure for the content of each tab


bggolding:
My question would be why do we forbid editing goals which have empty content? That makes no sense to me.

RaghavDabra:
@bggolding 

That’s a good point. After looking into the code, the restriction seems to come from `canEditGoal()` in **src/components/utils/GoalHint.tsx:**

```
export function isEmptyGoal(goal: { content: string }): boolean {
  return goal.content.trim() === "";
}
```
```

export function canEditGoal(goal: { content: string }): boolean {
  return !isEmptyGoal(goal);
}

```
Because `canEditGoal()` returns false when content is empty, goals with content: "" cannot be edited in the Arrange Hierarchy view.

_Initially I thought one possible fix would be to assign a default title (e.g. "New Goal") when creating goals via drag-and-drop in newTreeGoal() (src/components/types.ts)._

However, based on your point, it probably makes more sense to simply allow editing even when the content is empty. In that case, removing the restriction in `canEditGoal()` (e.g. making it always return true) would allow users to immediately enter a title after dragging a goal into the hierarchy.

bggolding:
I think there is an error in the code: editing a goal should be allowed even when the content is currently empty. That said, I don't think the user should be allowed to finish editing the goal content leaving it empty; that may be a separate issue though.

In your code, please do not put spaces inside braces, ie, instead of `canEditGoal(goal: { content: string })` prefer `canEditGoal(goal: {content: string})`. It makes the code more concise. You may need to update your IDE settings.

bggolding:
I've been having a good look at this. I think the problem needs to be resolved in Tree.tsx in `renderItem` (line ~304).

The code here isn't good. It renders the item differently depending on whether the item is being edited. There are a whole lot of tests based around the `isEditing` flag, turning lots of css and components on and off based on that. It shouldn't do this, instead there should be test at a higher level for `isEditing` and each branch of that should render a whole component as appropriate.

I think it should be changed to allow updating of the goal text even if it's currently empty. There are other checks in the code that prevent the user from submitting an empty item.

Maybe start by having a look at the code and then we can talk further.

RaghavDabra:
Hi Ben, Thanks for the detailed feedback. I’ve been working on the AMMBER reports and preparing a presentation for the mentor meeting, and then the software project items came in. I’m planning to review the Tree.tsx code and work on the suggested changes today. I’ll update you once I’ve had a closer look.

bggolding:
[I put this into our slack channel but it should be here]
For example, one of the things I noticed immediately was that the input fields are built manually inline rather than using some of the react bootstrap components that more naturally sit in the site’s style. I’m thinking specifically of [InputGroup](https://react-bootstrap.github.io/docs/forms/input-group#button-addons)

PR #262, opened 2026-03-08
fix: make zoom button group responsive

### LINKED ISSUE #221: Zoom toolbar exceeds background when window width is small
When the browser window is reduced to a certain width, the zoom toolbar exceeds the toolbar background container.
Image

Expected behavior:
The toolbar should remain within its background container at all window sizes. This can be fixed by setting a minimum width for the toolbar.

Current behavior:
The zoom toolbar overflows and extends beyond the toolbar background when the browser window becomes too narrow.

KerrYuan:
# Description
#221
[git issue](https://github.com/MotivationalModelling/mm-local-editor/issues/221)
Make the zoom button group fill parent width and allow buttons to wrap when space is too small.

# Current 
https://github.com/user-attachments/assets/99b4b0c1-82e8-42cd-a15e-24877421b7c2

### DIFF
--- a/src/components/Graphs/SidebarComponents/ZoomButtons.tsx
+++ b/src/components/Graphs/SidebarComponents/ZoomButtons.tsx
@@ -11,18 +11,21 @@ const ZoomButtons = ({recentreView}: ZoomButtonsProps) => {
     const {graph} = useGraph();
 
     return (
-        <ButtonGroup className="w-100" size="sm">
-            <Button variant="light"
+        <ButtonGroup className="w-100 d-flex flex-wrap" size="sm">
+            <Button className="flex-fill"
+                    variant="light"
                     size="sm"
                     onClick={() => graph?.zoomIn()}>
                 <BsZoomIn/>
             </Button>
-            <Button variant="light"
+            <Button className="flex-fill"
+                    variant="light"
                     size="sm"
                     onClick={() => recentreView()}>
                 <BsStopCircle/>
             </Button>
-            <Button variant="light"
+            <Button className="flex-fill"
+                    variant="light"
                     size="sm"
                     onClick={() => graph?.zoomOut()}>
                 <BsZoomOut/>
--- a/src/components/Graphs/SidebarComponents/ZoomButtons.tsx
+++ b/src/components/Graphs/SidebarComponents/ZoomButtons.tsx
@@ -25,7 +25,7 @@ const ZoomButtons = ({recentreView}: ZoomButtonsProps) => {
                 resizeObserver.disconnect();
             };
         }
-    }, [containerRef.current?.offsetWidth]);
+    }, []);
 
     return (
         <ButtonGroup ref={containerRef} 
--- a/src/components/Graphs/SidebarComponents/ZoomButtons.tsx
+++ b/src/components/Graphs/SidebarComponents/ZoomButtons.tsx
@@ -2,7 +2,7 @@ import ButtonGroup from "react-bootstrap/ButtonGroup";
 import Button from "react-bootstrap/Button";
 import {BsStopCircle, BsZoomIn, BsZoomOut} from "react-icons/bs";
 import {useGraph} from "../../context/GraphContext.tsx";
-import { useEffect, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


bggolding:
I don't think that simply using flex wrap is a good user experience because of the case where there are two buttons on one row and one on the next. I'd rather see the buttons switch from horizontal to vertical when the width hits some value.

I had a go at doing something like that using this:
 
```typescript
const ZoomButtons = ({recentreView}: ZoomButtonsProps) => {
    const {graph} = useGraph();
    const HORIZONTAL_THRESHOLD = 95;    // determined empirically
    const containerRef = useRef<HTMLDivElement>(null);
    const [containerWidth, setContainerWidth] = useState(0);

    useEffect(() => { // this runs on every render
        if (containerRef.current) {
            setContainerWidth(containerRef.current.offsetWidth);
        }
    });

    return (
        <ButtonGroup ref={containerRef}
                     className={`w-100 d-flex ${(containerWidth < HORIZONTAL_THRESHOLD) ? "flex-column" : "flex-row"}`}
                     size="sm">
        [...]
    );
}
```

KerrYuan:
@bggolding pls have a look again : P

src/components/Graphs/SidebarComponents/ZoomButtons.tsx
```diff
@@ -2,16 +2,37 @@ import ButtonGroup from "react-bootstrap/ButtonGroup";
 import Button from "react-bootstrap/Button";
 import {BsStopCircle, BsZoomIn, BsZoomOut} from "react-icons/bs";
 import {useGraph} from "../../context/GraphContext.tsx";
+import { useEffect, useRef, useState } from "react";
```
bggolding: Please do not put spaces inside braces. You might need to update your IDE settings.

src/components/Graphs/SidebarComponents/ZoomButtons.tsx
```diff
@@ -2,16 +2,37 @@ import ButtonGroup from "react-bootstrap/ButtonGroup";
...
     const {graph} = useGraph();
+    const HORIZONTAL_THRESHOLD = 90; 
+    const containerRef = useRef<HTMLDivElement>(null);
+    const [containerWidth, setContainerWidth] = useState(0);
+  
+    useEffect(() => {
+        const container = containerRef.current;
+
+        if (container) {
+            const resizeObserver = new ResizeObserver(() => {
```
bggolding, line 19: Very good -- this is better than rechecking on each render.

src/components/Graphs/SidebarComponents/ZoomButtons.tsx
```diff
@@ -2,16 +2,37 @@ import ButtonGroup from "react-bootstrap/ButtonGroup";
...
+            return () => {
+                resizeObserver.disconnect();
+            };
+        }
+    }, []);
 
     return (
-        <ButtonGroup className="w-100 d-flex flex-wrap" size="sm">
+        <ButtonGroup ref={containerRef} 
+                    className={`w-100 d-flex ${containerWidth < HORIZONTAL_THRESHOLD ? "flex-column" : "flex-row"}`}
```
bggolding: I always prefer to wrap the condition on `? :` expressions in parentheses so that unexpected operator precedence issues don't happen. It also makes it clear that this a condition that's being tested (similar to `if` and `while` statements).

bggolding (changes requested):

src/components/Graphs/SidebarComponents/ZoomButtons.tsx
bggolding: I don't think there is any value in `const container = containerRef.current;`. I think you should just remove `container` and replace it with `containerRef.current`.

In the dependency list of the `useEffect`, instead of it being an empty list, you could replace it with `..., [containerRef.current.offsetWidth]);`

bggolding (changes requested):
I just would like that small change to the `useEffect` dependencies made and then I'll merge the code.

bggolding (approved):
That looks great. Thanks Kerr.

PR #263, opened 2026-03-09
feature: allow user to show/hide line and keep distance

### LINKED ISSUE #238: Add the ability to turn the dot line on/off
It would be great to add a button in the toolbar to control whether the dot line connecting non-functional goals and functional goals appears in the graph or not.
Image

ToriLi1210:


src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -0,0 +1,85 @@
+import { useState } from "react";
+import { useFileContext } from "../../context/FileProvider.tsx";
+import { updateVisibilityForEdge } from "../../context/treeDataSlice.ts";
+import { ToggleButton, ToggleButtonGroup } from "react-bootstrap";
+
+
+const LineButtons = () => {
+    const [nonFunctionalExist, setNonFunctionalExist] = useState<number>(1);
```
bggolding: I don't think `nonFunctionalExist` is a good name because I don't understand what it means.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -538,8 +538,16 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({ showGraphSectio
...
+        const allEdges = graph.getChildEdges(graph.getDefaultParent());
+        allEdges.forEach((edge) => {
+            const style = graph.getCellStyle(edge);
+            const isNF = !!style?.dashed; 
+            edge.visible = isNF ? showNonFunctionalLines : showFunctionalLines;
+        });
+
         graph.getDataModel().endUpdate();
-    }, [graph, cluster]);
+    }, [graph, cluster,showNonFunctionalLines,showFunctionalLines]);
```
bggolding: Spaces after commas please.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -56,7 +56,7 @@ interface CellHistory {
 
 const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({ showGraphSection = false }) => {
     const divGraph = useRef<HTMLDivElement>(null);
-    const { cluster, dispatch, treeIds } = useFileContext();
+    const { cluster, dispatch, treeIds , showFunctionalLines,showNonFunctionalLines } = useFileContext();
```
bggolding: Missing space after comma.

bggolding: Remove spaces inside braces.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const [functionalExist, setFunctionalExist] = useState<number>(1);
+    const { dispatch } = useFileContext();
+
+    const changeVisibility = (isNonFunctionalEdge: boolean, val: number) => {
+        const isVisibile = val === 1;
+        dispatch(updateVisibilityForEdge({isNonFunctionalEdge, isVisibile}));
+    };
+
+    const miniBtnStyle = {
```
bggolding: I don't like the use of the style sheet overrides because it stops this button fitting into the bootstrap style. You should be able to get the `paddingLeft` and `paddingRight` by using `className="px-0"` in the `className` applied to the component. Changing the size seems unnecessary.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { updateVisibilityForEdge } from "../../context/treeDataSlice.ts";
+import { ToggleButton, ToggleButtonGroup } from "react-bootstrap";
+
+
+const LineButtons = () => {
+    const [nonFunctionalExist, setNonFunctionalExist] = useState<number>(1);
+    const [functionalExist, setFunctionalExist] = useState<number>(1);
+    const { dispatch } = useFileContext();
+
+    const changeVisibility = (isNonFunctionalEdge: boolean, val: number) => {
```
bggolding: I don't understand why this function is necessary. Why wouldn't you use:
```typescript
onChange={(val: number) => dispatch(updateVisibilityForEdge({isNonFunctionalEdge, isVisible: val === 1})
```
The thing is, because I can't understand what `nonFunctionalEdge` is, I can't understand what your code is doing.

src/components/Graphs/SidebarComponents/SidebarBody.tsx
```diff
@@ -24,6 +25,9 @@ const SidebarBody = ({graph, recentreView, className}: SidebarProps) => {
             <CollapsibleSidebarCard title="Font size">
                 <ScaleTextButton/>
             </CollapsibleSidebarCard>
+            <CollapsibleSidebarCard title="Line Visibility">
```
bggolding: Lowercase 'v' in visibility so that it's consistent with other `CollapsibleSideBar` headings.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -538,8 +538,16 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({ showGraphSectio
             qualitiesGlob,
             stakeholdersGlob
         );
+
+        const allEdges = graph.getChildEdges(graph.getDefaultParent());
+        allEdges.forEach((edge) => {
+            const style = graph.getCellStyle(edge);
+            const isNF = !!style?.dashed;
```
bggolding: I am really confused. It looks like you can select to show functional or non-functional lines between components but here it looks like they are just opposites. If they are just opposites, then just have a single value in the state.

This may be a result of the fact that names aren't clear.

I don't like that you are looking at how the lines are being rendered in the graph to as the criteria for whether to apply the changes. You should be updating something in the model or in the settings that render the graph instead. Using result of the rendering seems really bad to me.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -0,0 +1,85 @@
...
+                        id="nf-hide" value={0} variant="outline-secondary" 
+                        size="sm" className="flex-fill"
+                        style={miniBtnStyle}
+                    >
+                        Hide
+                    </ToggleButton>
+                </ToggleButtonGroup>
+            </div>
+
+            <div className="d-flex flex-column mb-3">
```
bggolding: Do you really need to use flex layout here? The default layout should be fine.

bggolding (changes requested):

src/components/context/FileProvider.tsx
```diff
@@ -91,8 +91,7 @@ interface FileContextProps {
     tabs: Map<Label, TabContent>
     goals: Record<TreeGoal["id"], TreeGoal>
     treeIds: Record<TreeGoal["id"], InstanceId[]>
-    showNonFunctionalLines: boolean
-    showFunctionalLines: boolean
+    showLineBetweenNonFunctionalGoal: boolean
```
bggolding: This is much better naming though I would probably suggest making it `showLinesBetweenNonFunctionalGoals` which makes it read better in English.

src/components/context/treeDataSlice.ts
```diff
@@ -261,15 +259,10 @@ export const treeDataSlice = createSlice({
                 node.color = color;
             }
         },
-        updateVisibilityForEdge: (state, action: PayloadAction<{
-            isNonFunctionalEdge: boolean,
+        updateVisibilityForLineBetweenNonFunctionalGoal: (state, action: PayloadAction<{
```
bggolding: Instead of `updateVisibilityForLineBetweenNonFunctionalGoal`, I would call this action `setVisibilityForLinesBetweenNonFunctionalGoals`.

src/components/context/treeDataSlice.ts
```diff
@@ -261,15 +259,10 @@ export const treeDataSlice = createSlice({
...
             isVisibile:boolean
```
bggolding: Instead of `isVisibile`, I would make this `showLines`. It makes it clear what this flag will update.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
 import { useFileContext } from "../../context/FileProvider.tsx";
-import { updateVisibilityForEdge } from "../../context/treeDataSlice.ts";
+import { updateVisibilityForLineBetweenNonFunctionalGoal } from "../../context/treeDataSlice.ts";
 import { ToggleButton, ToggleButtonGroup } from "react-bootstrap";
 
-
 const LineButtons = () => {
-    const [nonFunctionalExist, setNonFunctionalExist] = useState<number>(1);
-    const [functionalExist, setFunctionalExist] = useState<number>(1);
+    const [isVisible, setIsVisible] = useState<boolean>(true);
```
bggolding: I'm not sure why you've created a separate `useState` value for this. Instead of this you could pull this value from the FileContext using:
```
    const {showLinesBetweenNonFunctionalGoals} = useFileContext();
```

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
+                    className="w-100"
                     type="radio"
-                    name="functional-toggle"
-                    value={functionalExist}
-                    onChange={(val: number) => {
-                        setFunctionalExist(val);
-                        changeVisibility(false, val);
-                    }}
+                    name="nf-toggle"
+                    value={isVisible ? 1 : 0}
```
bggolding: If you pulled `showLinesBetweenNonFunctionalGoals` from the FileContext, you could use that here instead of `isVisible`.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
+                    onChange={handleChange}
                 >
-                    <ToggleButton 
-                        id="f-show" value={1} variant="outline-primary" 
-                        size="sm" className="flex-fill" 
-                        style={miniBtnStyle}
+                    <ToggleButton
+                        id="nf-show" 
+                        value={1}
```
bggolding: Ditto, if you use `showLinesBetweenNonFunctionalGoals` from the context, this will track its current setting rather than reverting to "1".

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
+                    name="nf-toggle"
+                    value={isVisible ? 1 : 0}
+                    onChange={handleChange}
                 >
-                    <ToggleButton 
-                        id="f-show" value={1} variant="outline-primary" 
-                        size="sm" className="flex-fill" 
-                        style={miniBtnStyle}
+                    <ToggleButton
+                        id="nf-show"
```
bggolding: Do we need an `id` attribute here? This component isn't likely to be picked out of the DOM by a selector so maybe remove this so that there's no confusion.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
                         Show
                     </ToggleButton>
-                    <ToggleButton 
-                        id="f-hide" value={0} variant="outline-secondary" 
-                        size="sm" className="flex-fill" 
-                        style={miniBtnStyle}
+                    <ToggleButton
+                        id="nf-hide" 
+                        value={0} 
+                        variant={!isVisible ? "primary" : "outline-secondary"}
```
bggolding: I'm a bit confused about why you would change the `variant` here? Shouldn't it just be a plain `ToggleButton`?

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
+                        className="p-0"
                     >
                         Show
                     </ToggleButton>
-                    <ToggleButton 
-                        id="f-hide" value={0} variant="outline-secondary" 
-                        size="sm" className="flex-fill" 
-                        style={miniBtnStyle}
+                    <ToggleButton
+                        id="nf-hide"
```
bggolding: Again, do we need `id`?

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
                     >
                         Show
                     </ToggleButton>
-                    <ToggleButton 
-                        id="f-hide" value={0} variant="outline-secondary" 
-                        size="sm" className="flex-fill" 
-                        style={miniBtnStyle}
+                    <ToggleButton
+                        id="nf-hide" 
+                        value={0}
```
bggolding: Should this use `showLinesBetweenNonFunctionalGoals`?

src/components/Graphs/SidebarComponents/SidebarBody.tsx
```diff
@@ -25,7 +25,7 @@ const SidebarBody = ({graph, recentreView, className}: SidebarProps) => {
             <CollapsibleSidebarCard title="Font size">
                 <ScaleTextButton/>
             </CollapsibleSidebarCard>
-            <CollapsibleSidebarCard title="Line Visibility">
+            <CollapsibleSidebarCard isOpen title="Line visibility">
```
bggolding: In the interests of keeping the UI cleaner, I wouldn't have this open by default. This might be a question to ask Leon in the next meeting though.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -56,7 +56,7 @@ interface CellHistory {
 
 const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({ showGraphSection = false }) => {
     const divGraph = useRef<HTMLDivElement>(null);
-    const { cluster, dispatch, treeIds , showFunctionalLines,showNonFunctionalLines } = useFileContext();
+    const { cluster, dispatch, treeIds, showLineBetweenNonFunctionalGoal } = useFileContext();
```
bggolding: No spaces inside braces please.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -541,13 +541,16 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({ showGraphSectio
 
         const allEdges = graph.getChildEdges(graph.getDefaultParent());
         allEdges.forEach((edge) => {
-            const style = graph.getCellStyle(edge);
-            const isNF = !!style?.dashed; 
-            edge.visible = isNF ? showNonFunctionalLines : showFunctionalLines;
-        });
+            const targetId = edge.target?.getId() || "";
```
bggolding: There must be a better way of doing this. We're looking at the rendered output and using that to determine which lines need to have their rendering changed. What I think we should be doing is updating the state that the rendering engine uses to draw the diagram and then ask it to re-render the diagram with the updated state.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
-                    >
-                        Hide
-                    </ToggleButton>
-                </ToggleButtonGroup>
-            </div>
-
             <div className="d-flex flex-column mb-3">
-                <span className="fw-bold mb-1" style={{ fontSize: '0.55rem' }}>Functional:</span>
+                <span className="fw-bold mb-1" style={{ fontSize: '0.7rem'}}>
+                    Between Non-Functioanl Goal
```
bggolding: There's a typo here: `Non-Functional`.

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
-                        style={miniBtnStyle}
-                    >
-                        Hide
-                    </ToggleButton>
-                </ToggleButtonGroup>
-            </div>
-
             <div className="d-flex flex-column mb-3">
-                <span className="fw-bold mb-1" style={{ fontSize: '0.55rem' }}>Functional:</span>
+                <span className="fw-bold mb-1" style={{ fontSize: '0.7rem'}}>
```
bggolding: Do we need to change the font size?

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -1,79 +1,46 @@
...
-    const changeVisibility = (isNonFunctionalEdge: boolean, val: number) => {
-        const isVisibile = val === 1;
-        dispatch(updateVisibilityForEdge({isNonFunctionalEdge, isVisibile}));
-    };
-
-    const miniBtnStyle = { 
-        fontSize: '0.55rem', 
-        paddingLeft: 0, 
-        paddingRight: 0 
+    const handleChange = (val: number) => {
```
bggolding: If we use `showLinesBetweenNonFunctionalGoals` from the `FileContext`, we should be able to remove this function because when the `dispatch(...)` happens, it will trigger a render.

src/components/context/FileProvider.tsx
bggolding: After reading further, do we need to have a `showLinesBetweenFunctionalGoals` as well?

bggolding (changes requested):
I'm sorry, there are a lot of comments here.

src/components/context/FileProvider.tsx
ToriLi1210: No, there should be lines between functional goals other the hierarchy will not be shown

src/components/Graphs/SidebarComponents/LineButtons.tsx
ToriLi1210: id is a compulsory parameter

src/components/Graphs/SidebarComponents/LineButtons.tsx
ToriLi1210: yes

src/components/Graphs/SidebarComponents/LineButtons.tsx
ToriLi1210: yes

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -0,0 +1,49 @@
+import {useFileContext} from "../../context/FileProvider.tsx";
+import {setVisibilityForLinesBetweenNonFunctionalGoals} from "../../context/treeDataSlice.ts";
+import {ToggleButton, ToggleButtonGroup} from "react-bootstrap";
+
+const LineButtons = () => {
+    const {showLineBetweenNonFunctionalGoal, dispatch} = useFileContext();
+    const isVisible = showLineBetweenNonFunctionalGoal;
```
bggolding: I had a look at this and I think we can make the whole body of this component much simpler:
```typescript
            <Form.Switch label="Between non-functional goals"
                         checked={showLineBetweenNonFunctionalGoal}
                         onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
                             dispatch(setVisibilityForLinesBetweenNonFunctionalGoals({
                                 showLines: e.target.checked
                             }));
                         }}/>
```

bggolding: I also don't think we really gain anything by using a named argument to `setVisibilityForLinesBetweenNonFunctionalGoals` so I would drop the `showLines:`. I have put a separate comment in treeDataSlice where that needs to happen.

src/components/context/FileProvider.tsx
```diff
@@ -91,6 +91,7 @@ interface FileContextProps {
     tabs: Map<Label, TabContent>
     goals: Record<TreeGoal["id"], TreeGoal>
     treeIds: Record<TreeGoal["id"], InstanceId[]>
+    showLineBetweenNonFunctionalGoal: boolean
```
bggolding: Can we just make this `showLineBetweenNonFunctionalGoals` (ie, add an 's')? It doesn't sound quite right in English having it singular.

src/components/context/treeDataSlice.ts
```diff
@@ -256,6 +259,11 @@ export const treeDataSlice = createSlice({
                 node.color = color;
             }
         },
+        setVisibilityForLinesBetweenNonFunctionalGoals: (state, action: PayloadAction<{
```
bggolding: We should make this:
```typescript
 setVisibilityForLinesBetweenNonFunctionalGoals: (state, action: PayloadAction<boolean>) => {
     state.showLineBetweenNonFunctionalGoal = action.payload;
},
```

src/components/Graphs/SidebarComponents/LineButtons.tsx
```diff
@@ -0,0 +1,49 @@
+import {useFileContext} from "../../context/FileProvider.tsx";
+import {setVisibilityForLinesBetweenNonFunctionalGoals} from "../../context/treeDataSlice.ts";
+import {ToggleButton, ToggleButtonGroup} from "react-bootstrap";
+
+const LineButtons = () => {
+    const {showLineBetweenNonFunctionalGoal, dispatch} = useFileContext();
+    const isVisible = showLineBetweenNonFunctionalGoal; 
+
+    return (
+        <div className="w-100 px-1">
```
bggolding: I did a test and I'm not sure that this `div` makes any difference so I'd remove it.

src/components/Graphs/GraphHelpers.tsx
```diff
@@ -392,7 +392,8 @@ export const renderNonFunction = (
     graph: Graph,
     source: Cell | null = null,
     type: string = "None",
-    color: string | undefined
+    color: string | undefined,
+    showLineBetweenNonFunctionalGoal: boolean
```
bggolding: Rather than pass this in as a parameter here, I would pull it out of the context using:
```typescript
const {showLineBetweenNonFunctionalGoals} = useFileContext();
```

src/components/Graphs/GraphHelpers.tsx
```diff
@@ -496,10 +497,12 @@ export const renderNonFunction = (
     );
     console.log("Nonfunctional-goal-node:",node);
     // Insert an invisible edge
-    const edge = graph.insertEdge(null, null, "", source, node, dotted);
-    // Ensure the line is behind the shape (Z-Index / Order)
-    // This moves the edge to the back of the graph display list
-    graph.orderCells(true, [edge]); 
+    if(showLineBetweenNonFunctionalGoal){
```
bggolding: Please put a space after the `if`.

src/components/Graphs/GraphHelpers.tsx
```diff
@@ -620,7 +623,8 @@ export const associateNonFunctions = (
     emotionsGlob: GlobObject,
     negativesGlob: GlobObject,
     qualitiesGlob: GlobObject,
-    stakeholdersGlob: GlobObject
+    stakeholdersGlob: GlobObject,
+    showLineBetweenNonFunctionalGoal: boolean
```
bggolding: By pulling `showLineBetweenNonFunctionalGoal` from the context where we use it, we can remove all these references to it.

bggolding (changes requested):
You have done outstanding work here. I love how you've added good tests and changed the way that the flag is used to do the rendering.

The few changes I've suggested here just tighten up the code and the UI a bit.

src/components/Graphs/GraphHelpers.tsx
ToriLi1210: I agree that using context is cleaner, but since renderNonFunction is a regular utility function used by the graph engine (not a React Functional Component), I can't call useFileContext directly inside it due to the Rules of Hooks. I'll pass it down as a parameter from the component level instead

bggolding: Yes, fair enough.

bggolding (approved):

PR #264, opened 2026-03-17
When adding a goal using drag and drop, the title isn't editable. #254

RaghavDabra:
Sorry for the late response on this bug: 
#254 

Goals dropped from the symbol palette were created with content: "" but the Tree component blocked editing any goal with empty content, making newly dropped goals permanently unnameable from the hierarchy view.

Three issues:

1. Remove isEmptyGoal guard from handleEdit/handleDoubleClick — the guard was intended to prevent editing deleted/phantom nodes but also fired for legitimately new empty goals. Saving an empty string is still blocked downstream by canSaveContentEdit so there is no regression.

2. Add useEffect that watches treeData for newly added nodes with empty content (detected via knownInstanceIds ref) and immediately enters edit mode with the input focused — consistent with standard tree editor UX.

3. On cancel/blur with no content entered, delete the orphan node from both the hierarchy and the goal list via deleteGoalReferenceFromHierarchy
   + deleteGoalFromGoalList so no permanently unnamed nodes are left behind.

Also fix dead-code logic bug in GoalHint.canSaveContentEdit where the second clause of the OR was unreachable (simplified to !isTextEmpty(newContent)).

Files changed:
  src/components/Tree.tsx
  src/components/utils/GoalHint.tsx

### DIFF
--- a/src/components/Tree.tsx
+++ b/src/components/Tree.tsx
@@ -1,4 +1,5 @@
-import React, {useRef, useState} from "react";
+// FIX #254: added useEffect to the import for auto-edit mode on new drag-and-drop goals
+import React, {useEffect, useRef, useState} from "react";
 import WhoIcon from "/img/Stakeholder.png";
 import DoIcon from "/img/Function.png";
 import BeIcon from "/img/Cloud.png";
@@ -19,7 +20,9 @@ import {
 } from "./utils/GoalHint.tsx"
 
 import "./Tree.css";
-import {deleteGoalReferenceFromHierarchy, setTreeData} from "./context/treeDataSlice.ts";
+// FIX #254: added deleteGoalFromGoalList so we can clean up orphan goals
+// that were drag-and-dropped but abandoned without a name
+import {deleteGoalFromGoalList, deleteGoalReferenceFromHierarchy, setTreeData} from "./context/treeDataSlice.ts";
 
 // Inline style for element in Nestable, css style import not working
 const treeListStyle: React.CSSProperties = {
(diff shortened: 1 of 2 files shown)


src/components/utils/GoalHint.tsx
```diff
@@ -7,16 +7,25 @@ export function isTextEmpty(input: string): boolean {
 }
 
 // Check if editing should be allowed (prevent editing empty goals)
+// FIX #254: this function is no longer called in Tree.tsx to gate handleEdit/handleDoubleClick —
```
bggolding, line 10: Do not include comments that aren't relevant to the code as it currently is; this refers to how the code used to be. As the project is developed, these comments all add up making it increasingly difficult to work out which comments refer to the current state of the project and which ones are "archaeology".

To be honest, I'm trying to think why this function even exists. Is there any situation where the goal's description should not be editable? Why would we ever want to stop a user doing that? We do need to make sure that if the user does edit the goal then there are criteria about what can be entered (eg, text not empty, doesn't share the same description as another goal at the same level, etc) but this is a separate issue that doesn't need to be addressed here.

src/components/utils/GoalHint.tsx
```diff
@@ -7,16 +7,25 @@ export function isTextEmpty(input: string): boolean {
...
+// empty goals created via drag-and-drop must be editable immediately.
+// It is kept here because it is still used elsewhere (e.g. isGoalDraggable, SectionPanel).
 export function canEditGoal(goal: { content: string }): boolean {
   return !isEmptyGoal(goal);
 }
 
 // Validate if save should be allowed (prevent saving empty content to existing goals)
+// FIX #254: the original second clause was dead code —
```
bggolding, line 18: Again, don't describe what was here. Your fix is good and clear as it is.

Note that you should have updated the old comment which describes the former behaviour. In this case, I would have deleted it.

At some stage it would be good to add in the other check that I mentioned in the previous comment (no duplicate goal names at the same level in the hierarchy). That could go in as a placeholder comment here for when that piece of work does get done because it functions as a placeholder to note where that code will need to go.

src/components/Tree.tsx
```diff
@@ -19,7 +20,9 @@ import {
 } from "./utils/GoalHint.tsx"
 
 import "./Tree.css";
-import {deleteGoalReferenceFromHierarchy, setTreeData} from "./context/treeDataSlice.ts";
+// FIX #254: added deleteGoalFromGoalList so we can clean up orphan goals
```
bggolding, line 23: Bad comment.

src/components/Tree.tsx
```diff
@@ -286,9 +353,13 @@ const Tree: React.FC<TreeProps> = ({
...
               className={`tree-input ${isTextEmpty(editedText) ? "is-invalid" : ""}`}
               style={treeInputStyle}
+              // FIX #254: placeholder so the user knows to type a name for a newly dropped goal
+              placeholder="Enter goal name…"
             />
           ) : (
-            treeItem.content
+            // FIX #254: show a grey fallback if content is empty so the node is still
+            // visually distinct and the user knows it needs a name
+            treeItem.content || <em style={{color: "#999"}}>unnamed goal</em>
```
bggolding, line 362: If you are going to use colours to flag something's status, use the standard ones from the bootstrap [colour palette](https://getbootstrap.com/docs/5.3/utilities/colors/), eg, `className="secondary-emphasis"`.

src/components/Tree.tsx
```diff
@@ -94,9 +97,52 @@ const Tree: React.FC<TreeProps> = ({
...
+  const editingWasNewEmptyGoal = useRef<boolean>(false);
+
+  // FIX #254: tracks all instanceIds already present in the tree so we can detect
+  // newly added nodes (i.e. ones dropped from the palette since the component mounted).
+  const knownInstanceIds = useRef<Set<InstanceId>>(new Set());
+
   const inputRef = useRef<HTMLInputElement>(null);
   const {treeData, dispatch} = useFileContext();
 
+  // FIX #254: auto-enter edit mode when a new empty goal is dropped from the palette.
```
bggolding, line 112: Rather than do this, we should be capturing the drop event and using that to get the details of the new goal.

However, the bigger question is why do we need to use a ref to handle this at all? This code really needs restructuring to remove this unnecessary complexity. This was the part of the code that I wanted to look at before you started working on this.

bggolding (changes requested):
In your PR, you mentioned three issues. I don't think 2 (using useEffect to track new goals from drag and drop) is a good way to do this. I don't think 3 (deleting goals that have empty descriptions) is sensible either -- I often do things like put a bunch of goals and then fill in their names. This would stop me doing that.

I think we need to stop using ref's to access these values if we can.

src/components/Tree.tsx
RaghavDabra: I fully agree on the comments, As you suggested I'll remove the fixes at #254  -> updating the existing ones to just reflect the current state of the code and switch the fallback colour back to the Bootstrap class. (lines 112, 291, fixing the comment 22, 18, 10)

On point 2 (deleting empty goals on cancel) — that makes sense, I hadn't considered the case where someone drops several goals first and names them afterwards. I'll revert that behaviour.

On point 1 (the useEffect / ref approach) — I'd really value your input here before I rework it. You mentioned wanting to look at Tree.tsx before I started, so I want to make sure I restructure it the way you had in mind rather than guess. Would you be able to point me in the right direction on how you'd prefer to capture the drop event and pass the new goal details through? 
Happy to jump on a call or just go back and forth here if that's easier.

PR #266, opened 2026-03-25
Refactor: replace Nestable with dnd-kit tree

### LINKED ISSUE #184: When dragging an item across the hierarchy list, the dragged element and its preview position are not aligned.

lollyluan:
Replaced the Nestable-based tree in Tree.tsx with a dnd-kit tree implementation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
       "name": "mm-local-editor",
       "version": "2025.2",
       "dependencies": {
+        "@dnd-kit/core": "^6.3.1",
+        "@dnd-kit/sortable": "^10.0.0",
+        "@dnd-kit/utilities": "^3.2.2",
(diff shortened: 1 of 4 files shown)


bggolding:
I tried this out and I like how it works. I've also tried it with React 19 and it works there too so it gives us an upgrade path; note that Nestable doesn't work with React 19.

Have you tested whether the unit tests still work with the new version? If they do, I'm inclined to merge this in.

lollyluan:
> I tried this out and I like how it works. I've also tried it with React 19 and it works there too so it gives us an upgrade path; note that Nestable doesn't work with React 19.
> 
> Have you tested whether the unit tests still work with the new version? If they do, I'm inclined to merge this in.

Hi Ben,

I’ve run all the tests and everything passed!

PR #273, opened 2026-04-07
196 bug switching window causes ownersvgelement null error in graphworker

### LINKED ISSUE #196: Bug: Switching window causes ownerSVGElement null error in GraphWorker
### Description

On the current `develop` branch, returning to the browser window after being idle on the Default Model page causes a runtime error in GraphWorker related to ownerSVGElement being null.
This breaks the graph rendering and prevents the editor from functioning properly.
The issue appears to be triggered when the graph is reset before the underlying SVG DOM is fully available
Image

### Steps to Reproduce

1. Start the local project (npm run dev).
2. Navigate to Default Model page.
3. Stay on the page for a while (or let the tab become inactive).
4. Switch to another app/window.
5. Switch back to the browser window.
6. The error appears in console and blank page

### Expected Behavior
Switching away from the window and returning should not break the graph.

aoxiangx:
The bug could not be reliably reproduced — the crash is not 100% consistent across attempts. 
However, a code audit of GraphWorker revealed the underlying vulnerability: the CHANGE listener attached to GraphDataModel was never removed on cleanup. After graph.destroy() nulls out view.drawPane, any late-firing CHANGE event — whether triggered by React scheduling, the browser resuming a throttled tab, a ResizeObserver callback, or something else — would cause graph.refresh() to dereference the nulled drawPane and crash with Cannot read properties of null (reading 'ownerSVGElement'). The fix extracts the handler into a named function, returns a cleanup callback from graphListener, and calls it before graph.destroy() in the effect cleanup. This ensures the listener is always removed before the graph's internal state is torn down, eliminating the race condition at its source.


bggolding (approved):
That looks really good Sean. Thanks!

PR #274, opened 2026-04-08
Changes for Tree.tsx

### LINKED ISSUE #254: When adding a goal using drag and drop, the title isn't editable.

RaghavDabra:
removed isEmptyGoal guard in handleEdit, removed isEmptyGoal check in handleDoubleClick, added unnamed goal fallback, removed disabled opacity/cursor style from edit icon, updated types from TreeItem to TreeGoal

### DIFF
--- a/src/components/SectionPanel.tsx
+++ b/src/components/SectionPanel.tsx
@@ -42,14 +42,12 @@ const DEFAULT_HEIGHT = "800px";
 type SectionPanelProps = {
   showGoalSection: boolean;
   showGraphSection: boolean;
-  setShowGoalSection: (showGoalSection: boolean) => void;
   paddingX: number;
 };
 
 const SectionPanel: React.FC<SectionPanelProps> = ({
   showGoalSection,
   showGraphSection,
-  setShowGoalSection,
   paddingX,
 }) => {
   const [sectionOneWidth, setSectionOneWidth] = useState(0);
@@ -189,11 +187,6 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
     setGroupSelected([]);
   };
 
-  // Handle synchronize data in table data and tree data
-  const handleSynTableTree = (treeItem: TreeGoal, editedText: string) => {
-    dispatch(updateTextForGoalId({id: treeItem.id, text: editedText}));
-  };
-
   // Get the parent div inner width and set starter width for section one and section three
   useEffect(() => {
     if (parentRef.current) {
@@ -284,15 +277,8 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
         onDragOver={(event) => event.preventDefault()}
         ref={sectionTwoRef}
       >
-        <Tree
-
-          // existingItemIds={existingItemIds}
-          // setTreeIds={setTreeIds}
-          handleSynTableTree={handleSynTableTree}
-          // setExistingItemIds={setExistingItemIds}
-          existingGoalReferenceInstanceId={existingGoalReferenceInstanceId}
-          setExistingGoalReferenceInstanceId={setExistingGoalReferenceInstanceId}
-        />
+          <Tree existingGoalReferenceInstanceId={existingGoalReferenceInstanceId}
+                setExistingGoalReferenceInstanceId={setExistingGoalReferenceInstanceId}/>
       </div>
 
       {/* Graph Render Section */}
--- a/src/components/SectionPanel.tsx
+++ b/src/components/SectionPanel.tsx
@@ -145,7 +145,7 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
 
       if (draggedItem && draggedItem.content) {
             // the first hierachy does not contain the dragged item
-            if (!tree.map((index) => index.id).includes(draggedItem.id)) {
+            if (!tree.some((goal) => goal.id === draggedItem.id)) {
               dispatch(addGoalToTree(draggedItem));
           } else {
               setExistingItemIds([...existingItemIds, draggedItem.id]);
@@ -162,7 +162,7 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
     const newItemsToAdd = groupSelected.filter(
             // current hierachy
             (item) => !tree.some(
-                ref => ref.id === item.id
+                goal => goal.id === item.id
             )
     );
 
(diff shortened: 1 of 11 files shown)


src/components/utils/GoalHint.tsx
```diff
@@ -11,12 +11,11 @@ export function canEditGoal(goal: { content: string }): boolean {
   return !isEmptyGoal(goal);
 }
 
-// Validate if save should be allowed (prevent saving empty content to existing goals)
+// Validate if save should be allowed (prevent saving empty content)
```
bggolding, line 14: It's good to see you updating the existing comment to make it accurate following your changes.

src/components/utils/GoalHint.tsx
```diff
@@ -11,12 +11,11 @@ export function canEditGoal(goal: { content: string }): boolean {
...
+// TODO: also validate that the goal name is not a duplicate at the same tree level
```
bggolding, line 15: And this is good too because when someone comes along later to make that change, this is a strong clue that this will be the right place to make that change.

These sorts of small changes to comments add significant value to the codebase.

src/components/Tree.tsx
```diff
@@ -288,7 +283,7 @@ const Tree: React.FC<TreeProps> = ({
               style={treeInputStyle}
             />
           ) : (
-            treeItem.content
+            treeItem.content || <em className="text-secondary">unnamed goal</em>
```
bggolding: Does this need to be an `em`? Perhaps a `span` might be more appropriate?

I would make the text `Goal name` instead of `unnamed goal` to tell the user precisely what the semantics of the field are.

src/components/context/treeDataSlice.ts
```diff
@@ -158,6 +158,17 @@ export const findTreeGoalByInstanceId = (nodes: TreeGoal[], instanceId: Instance
     return undefined;
 };
 
+const updateTreeGoalContentById = (nodes: TreeGoal[], id: TreeGoal["id"], text: string): void => {
+    for (const node of nodes) {
```
bggolding: use `nodes.forEach((node) => { ...`

src/components/context/treeDataSlice.ts
```diff
@@ -158,6 +158,17 @@ export const findTreeGoalByInstanceId = (nodes: TreeGoal[], instanceId: Instance
...
+        if (node.id === id) {
+            node.content = text;
+        }
+        if (node.children && node.children.length > 0) {
+            updateTreeGoalContentById(node.children, id, text);
```
bggolding: I think you are missing `dispatch`, ie,
```typescript
dispatch(updateTreeGoalContentById(node.children, id, text)));
```

src/components/context/treeDataSlice.ts
```diff
@@ -227,6 +238,7 @@ export const treeDataSlice = createSlice({
                 ...state.goals[action.payload.id],
                 content: action.payload.text
             };
+            updateTreeGoalContentById(state.tree, action.payload.id, action.payload.text);
```
bggolding: ditto `dispatch`

src/components/SectionPanel.tsx
```diff
@@ -144,7 +145,7 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
 
       if (draggedItem && draggedItem.content) {
             // the first hierachy does not contain the dragged item
-            if (!tree.map((index) => index.goalId).includes(draggedItem.id)) {
+            if (!tree.map((index) => index.id).includes(draggedItem.id)) {
```
bggolding: This looks really clumsy, some of this is because the naming is wrong: instead of `index`, it should be called `goal`. I'd be more inclined to write this like:
```typescript
if (tree.some((goal) => goal.id === draggedItem.id)) {
```
(I think that's right, double check it)

bggolding (changes requested):

src/components/context/treeDataSlice.ts
```diff
@@ -291,5 +276,5 @@ export const treeDataSlice = createSlice({
     }
 });
 
-export const {addGoal, addGoalToTab, setTreeData, addGoalToTree, deleteGoalReferenceFromHierarchy, deleteGoalFromGoalList, updateTextForGoalId, reset, removeGoalIdFromTree, updateTextForInstanceId, updateColorForInstanceId} = treeDataSlice.actions;
+export const {addGoal, addGoalToTab, setTreeData, addGoalToTree, deleteGoalReferenceFromHierarchy, deleteGoalFromGoalList, updateTextForGoalId, updateTextForInstanceId, reset, removeGoalIdFromTree, updateColorForInstanceId} = treeDataSlice.actions;
```
bggolding: Why did you update the order of these? They were in alphabetical order.

src/components/SectionPanel.tsx
```diff
@@ -162,7 +162,7 @@ const SectionPanel: React.FC<SectionPanelProps> = ({
     const newItemsToAdd = groupSelected.filter(
             // current hierachy
             (item) => !tree.some(
-                ref => ref.id === item.id
+                goal => goal.id === item.id
```
bggolding, line 162: Much better naming!

src/components/Tree.tsx
```diff
@@ -159,11 +159,12 @@ const Tree: React.FC<TreeProps> = ({
   const renderItem: NestableProps["renderItem"] = ({item, collapseIcon}) => {
     const treeItem = item as TreeGoal;
     const isEditing = editingItemId === treeItem.id;
+    const goalContent = goals[treeItem.id]?.content ?? "";
```
bggolding: This is a good name for this variable -- it's easy to remember where its value has come from.

bggolding (changes requested):
There are just a few minor changes to make here. Once they're done, I'll make a few changes on this branch that I talked about earlier. I'll ask you to review those changes when I've done them.

src/components/context/treeDataSlice.ts
RaghavDabra: Hi Ben, I didn’t realise the exports were alphabetically ordered. I’ve reverted them back to the original ordering I've done the commits.

PR #277, opened 2026-04-19
clear corrupted localStorage data on parse failure

### LINKED ISSUE #107: Bug/Enhancement: Clear invalid LocalStorage data on Zod parse failure
Currently, when loading data from LocalStorage, if the data is invalid (e.g., corrupted, outdated, or manually modified), the application tries to parse it but fails. However, the invalid data remains in LocalStorage, which leads to repeated Zod parse errors on every reload and prevents the app from recovering.

Steps to Reproduce:

Put invalid JSON or outdated schema data into LocalStorage under appData (or related keys).

Reload the app.

Observe: Zod parsing fails and app cannot recover because bad data persists.

aoxiangx:
JSON.parse threw on malformed data and the bad value stayed in localStorage, crashing the app on every reload. Now the key is removed on parse failure and the initial value is used instead.

Also fixes a logic bug where the Function form of initialValue bypassed stored data unconditionally.

Test plan

1. Open the app normally, add some goals, refresh — confirm data persists from localStorage
2. Open DevTools → Application → Local Storage, manually corrupt the value of ammber/treeData (e.g. change it to {{{invalid}}})
3. Refresh the page — confirm the app loads without crashing and starts with a clean initial state
4. Check localStorage again — confirm the corrupted key has been removed
5. Add goals again, refresh — confirm data persists normally after recovery


src/components/utils/useLocalStorage.tsx
```diff
@@ -6,27 +6,29 @@ type InitialValueType<T> = T | (() => T);
 const isEmptyObject = (obj: object) =>
   obj && typeof obj === "object" && Object.keys(obj).length === 0;
 
+const resolveInitialValue = <T,>(initialValue: InitialValueType<T>): T =>
+	initialValue instanceof Function ? initialValue() : initialValue;
```
bggolding: To be honest, I don't see there's any point in allowing initialValue to be a function. I would only accept a value here and if they want to use a function, they can call it when they set the value.

I feel like this is adding complexity and flexibility when it's not really needed.

src/components/utils/useLocalStorage.tsx
```diff
@@ -6,27 +6,29 @@ type InitialValueType<T> = T | (() => T);
...
-
-	if (initialValue instanceof Function) {
-		return initialValue();
-	}
-
-	if (Array.isArray(data) && data.length === 0) {
-		return initialValue;
+	if (storedData === null) return resolveInitialValue(initialValue);
+
+	let data: unknown;
```
bggolding: I am always wary of using `let` and try to avoid it wherever possible. I would try to move the calculation of `data` into a function and then initialise a `const` `data` from that.

I'm not sure but maybe removing acceptation a function might help solve this.

bggolding (changes requested):
I really like what you're trying to do here -- without this, a user would have to manually delete the broken data from localStorage in their browser, I just think we can do this more cleanly.

bggolding:
Looking further at this code, I really think we should remove `useLocalStorage.tsx` and use an external module instead. That means someone else maintains that code and the less code in the project, the better.

Maybe we could use this package [use-local-storage](https://www.npmjs.com/package/use-local-storage)? The interface looks largely identical.

aoxiangx:
2 new commits/fixes as promised.

**Commit 1 — Replace custom hook with use-local-storage package**

Deleted useLocalStorage.tsx and switched to the npm package [use-local-storage](https://www.npmjs.com/package/use-local-storage). The package exposes the same useState-shaped API and wraps JSON.parse in a try/catch with a logger fallback, so the corrupted-data recovery from the previous commit is preserved with no project-side hook to maintain. Both callers in FileProvider are unchanged structurally.

**Commit 2 — Drop tree when it references goals not present in tabData**

While verifying commit 1 end-to-end I corrupted only ammber/tabData and left ammber/treeData intact. The parse-failure recovery worked — tabData fell back to defaults — but the app still crashed, this time with Uncaught Error: goal X in tree but not in goal list thrown from createTreeIdsFromTreeData.

**_Root cause:_** when only one of the two storage keys recovers to defaults, the surviving key can still reference goal ids that no longer exist in the new tabData. The downstream invariant check is correct and worth keeping; the data has to be validated at the storage boundary instead.

**_Fix:_** at the boundary, check whether every node in treeData references a goal id present in tabData. If any node is dangling, drop the tree to []. The existing write-back effect rewrites the stale storage on the next render, so the app self-heals.

**How to verify**
1. Open the page, add a few goals / tabs so storage holds non-default data (or use default model instead).
2. DevTools → Application → Local Storage. Edit one or both of ammber/tabData and ammber/treeData to invalid values (e.g. not valid json {{{, {"broken":, undefined).
3. Refresh.

Expected:
- Page loads normally — 5 default tabs, empty tree if the cross-reference check tripped.
- Console shows a caught SyntaxError from the use-local-storage library logger. This is the package reporting that it handled the error, not an uncaught app error.
- No Uncaught Error: goal X in tree but not in goal list.
- The stale storage entry gets overwritten on the next render with valid JSON, so a second refresh is clean.

bggolding:
I was wondering why not wrap the call to `createInitialState` in a `try`/`catch` and if there is an exception thrown for whatever reason in that process, then tell the user that the state is corrupted and perhaps give them an option to abandon or zap the saved state?

aoxiangx:
> I was wondering why not wrap the call to `createInitialState` in a `try`/`catch` and if there is an exception thrown for whatever reason in that process, then tell the user that the state is corrupted and perhaps give them an option to abandon or zap the saved state?


Great idea — happy to go with that approach. A few things I'd like to align on before implementing:

**Abandon:** refresh the page and reset localStorage along with it — otherwise a plain refresh would just re-load the same corrupted data and we'd hit the same error again.

**Zap:** what do you have in mind here? Or is there a snapshot mechanism I'm missing where we could roll back to a pre-corruption version of the model?

**UI** — thinking a modal, styled to match the existing "delete goal" confirm popup so the look stays consistent.

**Scope** — once we agree, should this replace the cross-reference fix in `ee26cf38`, or sit on top of it as a second safety net?

bggolding:
The options I would offer are "abandon" and the app would just exit leaving the localstorage state unchanged as well as revert to default and just replace it with the default state.

aoxiangx:
> The options I would offer are "abandon" and the app would just exit leaving the localstorage state unchanged as well as revert to default and just replace it with the default state.

Just to confirm — for "abandon", do you want the app to render an empty editor state (5 tabs, no goals) where the user can keep working, or to not render the editor at all and leave the page in some kind of paused / recovery state? The first option means the corrupted localStorage gets overwritten on first interaction; the second preserves it.

One thing worth flagging on the second variant: if we truly leave localStorage unchanged, then reopening the app just re-triggers the same modal — the user's only real exits are picking "revert to default" or clearing localStorage manually in DevTools. Is that the intent? (Fine if so — it's effectively an escape hatch for technical users — just want to be sure.)

@bggolding ben

bggolding:
The second option is what I would like to see. It gives an expert user the possibility of dealing with the corruption.

aoxiangx:
## What changed
`FileProvider` now validates the saved model (`ammber/treeData`, `ammber/tabData`)
on load and shows a "Saved data is corrupted" modal instead of crashing. The two
keys are stored as raw JSON strings so corrupted data is neither swallowed nor
overwritten on load. Replaces the custom `useLocalStorage` hook with the
`use-local-storage` package. 

## Behavior
- Blank / default / absent data loads normally.
- Any unusable data — unparseable JSON, wrong shape, or a tree referencing a goal
  not in the tabs — shows the modal with two choices:
  - **Revert to default**: replace the saved data with the default state.
  - **Abandon**: pause and leave localStorage untouched for inspection.

## Testing
`npm test` — 66/66 pass. The `corrupted localStorage recovery` suite covers
unparseable JSON, wrong-shape tabs, missing-goal trees, Abandon leaving storage
byte-for-byte intact, and Revert restoring the editor.

image

PR #280, opened 2026-05-07
fix: resolve the problem when editing the tree item

lollyluan:


bggolding:
I can't merge this into the #254 branch which has been hugely refactored and reworked. Could you look at this again please using that branch.

lollyluan:
> I can't merge this into the #254 branch which has been hugely refactored and reworked. Could you look at this again please using that branch.

No problem, I will check that branch again and work base on that one.

bggolding:
Thanks

bggolding:
I am confused. Your changes don't appear on the #254 branch. Please make the changes there.

In the changes on the #184 branch, I don't understand why you need a useEffect.

lollyluan:
> I am confused. Your changes don't appear on the #254 branch. Please make the changes there.
> 
> In the changes on the #184 branch, I don't understand why you need a useEffect.

The bug was caused by the interaction between the tree re-render cycle and the DnD wrapper. Clicking a row first updates the edit state and triggers a re-render, but the input field does not exist until after that render completes. I used useEffect for selecting logic runs only after the component has entered edit mode and the input has been recorded.

lollyluan:
> I am confused. Your changes don't appear on the #254 branch. Please make the changes there.
> 
> In the changes on the #184 branch, I don't understand why you need a useEffect.

I have fixed the issue on the #184 branch based on #254 branch. Would you like to test on this branch, or do you like me to bring these changes to #254 branch?

src/components/context/treeDataSlice.ts
```diff
@@ -305,6 +246,7 @@ export const treeDataSlice = createSlice({
                 ...state.goals[action.payload.id],
                 content: action.payload.text
             };
+            updateTreeGoalTextById(state.tree, action.payload.id, action.payload.text);
```
bggolding: Are you sure that this call is needed? The goals are linked into `state.goals` as well as `state.tree` so updating a goal in one should update the same goal in the other.

src/components/Tree.tsx
```diff
@@ -1,105 +1,434 @@
...
+}, ref) => {
+    const treeItem = item as SortableTreeGoal;
+    const isEditing = editingItemId === treeItem.instanceId;
+    const iconSize = 25;
+    const [draftText, setDraftText] = useState(treeItem.content);
+    const isReference = existingGoalReferenceInstanceId.some(
+      (itemRef) => itemRef.goalId === treeItem.id && itemRef.instanceId === treeItem.instanceId
+    );
+
+    useEffect(() => {
```
bggolding: Rather than use `useEffect` here, I think you want to trigger `setDraftText(treeItem.content)` at the same time that you set `editingItemId` to `null`, even if you need to call a function in a couple of places. This will make the code much easier to understand than using `useEffect` because it connects the action that triggers `setDraftText` directly rather than it being triggered indirectly.

bggolding (changes requested):
What you've done is mostly good but I think there's a little room to make it better.

PR #282, opened 2026-07-19
feat: change concern shape to triangle

lvqq2000:
image
image


bggolding (changes requested):
There is no drop shadow on the new icon; all the other icons have them so it looks out of place.

When you are drawing the image, it has sharp edges and the "!" doesn't match the style of the png. I think it would be better to use an image for this rather than drawing. This would be consistent with the other icons when they're drawn in the graph.

Leon was going to talk to the person who had originally prepared the png images for the icon to see if they could do this one in the same style. I haven't heard whether that is going to proceed.

Please use the "create branch" on the ticket so that branches are created with consistent names. You have created this as a branch called "282" which is difficult to understand what it is doing.

PR #285, opened 2026-07-28
feat(211): Add related links to goals

### LINKED ISSUE #211: Link to related documents
When creating a model, it maybe useful to include links to files or web pages that are relevant to the goals. It would be good to have a mechanism where we can add those in and then be able to inspect them.

I think just having a url associated with the goal would probably be sufficient. When it was clicked on, we could just hand that to the browser to handle itself whether it's a file (file://), a link (https://) or something else.

ShengboYY:
## Summary

Adds support for associating related web links with goals.

## Changes

- Added a link button to each goal in the goal list.
- Added the same link action to goals in the hierarchy view.
- Added a modal for adding, updating, opening, and removing URLs.
- Restricted links to valid `http://` and `https://` URLs.
- Persisted goal URLs with the existing local goal data.
- Displayed linked goals as blue, underlined text in the rendered model.
- Made rendered goal text clickable without affecting node dragging or editing.
- Preserved clickable links when exporting the model as SVG.

### Link button
image
image

### Popup window
image

### Render model
image

### DIFF
--- a/src/components/GoalLinkModal.tsx
+++ b/src/components/GoalLinkModal.tsx
@@ -2,25 +2,22 @@ import React, {useState} from "react";
 import Button from "react-bootstrap/Button";
 import Form from "react-bootstrap/Form";
 import Modal from "react-bootstrap/Modal";
+import {useFileContext} from "./context/FileProvider";
+import {updateUrlForGoalId} from "./context/treeDataSlice.ts";
 import {TreeGoal} from "./types.ts";
 import {linkTitleForGoal} from "./linkTitleForGoal.ts";
 
 type GoalLinkModalProps = {
-	show: boolean;
-	goal: TreeGoal | null;
-	onHide: () => void;
-	onRemove: () => void;
-	onSave: (url: string) => void;
+	goal: TreeGoal;
+	onClose: () => void;
 };
 
 const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
-	show,
 	goal,
-	onHide,
-	onRemove,
-	onSave,
+	onClose,
 }) => {
-	const [url, setUrl] = useState(goal?.url ?? "");
+	const {dispatch} = useFileContext();
+	const [url, setUrl] = useState(goal.url ?? "");
 	const normalizedUrl = url.trim();
 
 	const isValidUrl = () => {
@@ -38,8 +35,14 @@ const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
 	const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
 		event.preventDefault();
 		if (isUrlValid) {
-			onSave(normalizedUrl);
+			dispatch(updateUrlForGoalId({id: goal.id, url: normalizedUrl}));
+			onClose();
 		}
+	};
+
+	const handleRemove = () => {
+		dispatch(updateUrlForGoalId({id: goal.id, url: undefined}));
+		onClose();
 	};
 
 	const handleOpen = () => {
@@ -49,7 +52,7 @@ const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
 	};
 
 	return (
-		<Modal show={show} onHide={onHide} centered>
+		<Modal show onHide={onClose} centered>
 			<Modal.Header closeButton>
 				<Modal.Title>{linkTitleForGoal(goal)}</Modal.Title>
 			</Modal.Header>
@@ -80,8 +83,8 @@ const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
 					>
 						Open link
 					</Button>
-					{goal?.url && (
-						<Button type="button" variant="outline-danger" onClick={onRemove}>
+					{goal.url && (
+						<Button type="button" variant="outline-danger" onClick={handleRemove}>
 							Remove
 						</Button>
 					)}
--- a/src/components/GoalLinkModal.tsx
+++ b/src/components/GoalLinkModal.tsx
@@ -54,7 +54,8 @@ const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
 	return (
 		<Modal show={show} onHide={onHide} centered>
 			<Modal.Header closeButton>
-				<Modal.Title>Related link</Modal.Title>
+				{/* Identify the goal whose link is being edited in the dialog title. */}
+				<Modal.Title>Link for {goal?.content || "this goal"}</Modal.Title>
 			</Modal.Header>
 			<Form onSubmit={handleSubmit}>
 				<Modal.Body>
@@ -71,9 +72,6 @@ const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
 						<Form.Control.Feedback type="invalid">
 							Enter a valid http:// or https:// URL.
 						</Form.Control.Feedback>
-						<Form.Text muted>
-							Link for {goal?.content || "this goal"}
-						</Form.Text>
 					</Form.Group>
 				</Modal.Body>
 				<Modal.Footer>
@@ -91,9 +89,6 @@ const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
 							Remove
 						</Button>
 					)}
-					<Button type="button" variant="secondary" onClick={onHide}>
-						Cancel
-					</Button>
 					<Button type="submit" variant="primary" disabled={!isUrlValid}>
 						Save
 					</Button>
--- a/src/components/GoalLinkModal.tsx
+++ b/src/components/GoalLinkModal.tsx
@@ -1,0 +1,106 @@
+import React, {useEffect, useState} from "react";
+import Button from "react-bootstrap/Button";
+import Form from "react-bootstrap/Form";
+import Modal from "react-bootstrap/Modal";
+import {TreeGoal} from "./types.ts";
+
+type GoalLinkModalProps = {
+	show: boolean;
+	goal: TreeGoal | null;
+	onHide: () => void;
+	onRemove: () => void;
+	onSave: (url: string) => void;
+};
+
+const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
+	show,
+	goal,
+	onHide,
+	onRemove,
+	onSave,
+}) => {
+	const [url, setUrl] = useState("");
+	const normalizedUrl = url.trim();
+
+	useEffect(() => {
+		setUrl(goal?.url ?? "");
+	}, [goal]);
+
+	const isValidUrl = () => {
+		try {
+			const parsedUrl = new URL(normalizedUrl);
+			return parsedUrl.protocol === "http:" || parsedUrl.protocol === "https:";
+		} catch {
+			return false;
+		}
+	};
+
+	const isUrlValid = isValidUrl();
+	const showInvalidUrl = url.length > 0 && !isUrlValid;
+
+	const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault();
+		if (isUrlValid) {
+			onSave(normalizedUrl);
+		}
+	};
+
+	const handleOpen = () => {
+		if (isUrlValid) {
+			window.open(normalizedUrl, "_blank", "noopener,noreferrer");
+		}
+	};
+
+	return (
+		<Modal show={show} onHide={onHide} centered>
+			<Modal.Header closeButton>
+				<Modal.Title>Related link</Modal.Title>
+			</Modal.Header>
+			<Form onSubmit={handleSubmit}>
+				<Modal.Body>
+					<Form.Group controlId="goal-related-link">
+						<Form.Label>URL</Form.Label>
+						<Form.Control
+							type="url"
+							value={url}
+							placeholder="Enter an http:// or https:// URL"
+							isInvalid={showInvalidUrl}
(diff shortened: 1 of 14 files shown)


src/components/GoalLinkModal.tsx
```diff
@@ -54,7 +54,8 @@ const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
 	return (
 		<Modal show={show} onHide={onHide} centered>
 			<Modal.Header closeButton>
-				<Modal.Title>Related link</Modal.Title>
+				{/* Identify the goal whose link is being edited in the dialog title. */}
+				<Modal.Title>Link for {goal?.content || "this goal"}</Modal.Title>
```
bggolding: I would factor this out so that it looks more like:
```jsx
<Modal.Title>{linkTitleForGoal(goal)}</Modal.Title>
```
You can then get rid of the comment because the function name describes the action, you can reuse the code, and it becomes (potentially) testable.

src/components/GoalLinkModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+	show,
+	goal,
+	onHide,
+	onRemove,
+	onSave,
+}) => {
+	const [url, setUrl] = useState("");
+	const normalizedUrl = url.trim();
+
+	useEffect(() => {
```
bggolding: You do not need this `useEffect`, instead just change the initialiser to:
```typescript
const [url, setUrl] = useState(goal.?url ?? "");
```

`useEffect` makes it very difficult to reason about flow control in code and you should try to use them as sparingly as possible.

src/components/Graphs/GraphWorker.tsx
```diff
@@ -272,6 +295,22 @@ const GraphWorker: React.FC<{ showGraphSection?: boolean }> = ({showGraphSection
 
     const graphListener = useCallback((graph: Graph): (() => void) => {
         const cellHistory: CellHistory = {};
+        const goalLinkClickHandler = (_sender: string, evt: EventObject) => {
+            const cell = evt.getProperty("cell") as Cell | null;
```
bggolding: Why not use:
```typescript
const cell = evt.getProperty("cell") as Cell ?? null;
```

src/components/Graphs/GraphHelpers.tsx
```diff
@@ -48,6 +50,26 @@ const FUNCTIONAL_NONFUNCTIONAL_OFFSET = {
...
+
+// Keep link metadata outside the cell value so existing plain-text editing
+// continues to work without HTML labels.
+export const getGoalLinkForCell = (cell: Cell): string | undefined => goalLinks.get(cell);
+
+const applyGoalLink = (style: CellStyle, url?: string) => {
+    if (!url) return;
+
+    style.fontColor = GOAL_LINK_COLOR;
+    style.fontStyle = (style.fontStyle ?? 0) | constants.FONT.UNDERLINE;
```
bggolding: I would not modify the object that's passed in, return a copy of the `style` with the extra attribute instead:
```typescript
return {
    ...style,
    fontColor: GOAL_LINK_COLOR,
    fontStyle: (style.fontStyle ?? 0) | constants.FONT.UNDERLINE;
};
```
An object that's passed in isn't "yours" to modify, you don't know how else it's used.

src/components/context/treeDataSlice.ts
```diff
@@ -231,6 +248,22 @@ export const treeDataSlice = createSlice({
                 content: action.payload.text
             };
         },
+        updateUrlForGoalId: (state, action: PayloadAction<{
+            id: TreeGoal["id"],
+            url?: string
+        }>) => {
+            const goal = state.goals[action.payload.id];
+            if (!goal) return;
```
bggolding: This is probably an exception, not just a `return`. Putting an exception here will help catch internal consistency errors quickly.

src/components/context/treeDataSlice.ts
```diff
@@ -117,6 +117,23 @@ const removeAllReferenceFromHierarchy = (
         }));
 };
 
+const updateUrlForGoalReferences = (
+    nodes: TreeGoal[],
+    goalId: TreeGoal["id"],
+    url?: string,
```
bggolding: Rather than make this an optional argument, I think you should make it:
```typescript
url: string | undefined
```
With that done, you can replace the body of the code with:
```typescript
if (node.id === goalId) {
    node.url = url;
}
```

src/components/context/treeDataSlice.ts
```diff
@@ -231,6 +248,22 @@ export const treeDataSlice = createSlice({
                 content: action.payload.text
             };
         },
+        updateUrlForGoalId: (state, action: PayloadAction<{
+            id: TreeGoal["id"],
+            url?: string
```
bggolding: Again, rather than making this an optional parameter, use:
```typescript
url: string | undefined
```

src/components/context/treeDataSlice.ts
```diff
@@ -231,6 +248,22 @@ export const treeDataSlice = createSlice({
...
+        }>) => {
+            const goal = state.goals[action.payload.id];
+            if (!goal) return;
+
+            if (action.payload.url) {
```
bggolding:
```typescript
if (action.payload.url) {
    goal.url = action.payload.url;
}
```

src/components/Graphs/GraphHelpers.tsx
```diff
@@ -108,13 +130,13 @@ export const renderGoals = (
 
             // accumulate non-functional descriptions into buckets
         } else if (type === SYMBOL_CONFIGS.EMOTIONAL.type) {
-            emotions.push({instanceId, content});
+            emotions.push({instanceId, content, url: goal.url});
```
bggolding, line 136: It's not clear to me why you need to pass a `url` parameter to this.

src/components/Graphs/GraphHelpers.tsx
```diff
@@ -455,6 +479,11 @@ export const renderNonFunction = (
     style.labelPosition = "center";
     style.spacingTop = 0;
 
+    // Aggregated symbols can represent several goals, so only create a text
+    // link when the rendered label has one unambiguous URL target.
+    const goalUrl = descriptions.length === 1 ? descriptions[0].url : undefined;
```
bggolding: I think it is good practice to wrap the conditional expression in a `?:` expression in parentheses:
```typescript
const goalUrl = (descriptions.length === 1) ? descriptions[0].url : undefined;
```

I don't understand the comment: how can a goal have multiple url's?

src/components/GoalLinkModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+type GoalLinkModalProps = {
+	show: boolean;
+	goal: TreeGoal | null;
+	onHide: () => void;
+	onRemove: () => void;
+	onSave: (url: string) => void;
+};
+
+const GoalLinkModal: React.FC<GoalLinkModalProps> = ({
```
bggolding, line 15: I really think you're being overly general here. This component is specifically for updating the link associated with a goal so just make it do that.

 Remove the `show` prop and only render it when it's going to be displayed:
```jsx
{(showGoalLink) && <GoalLinkModal goal={goal} .../>}
```

The only props it should need are: `goal` and `onClose`. Move the code that handles updating the `goal`'s url into this component so that it's not scattered through the codebase.

src/components/GoalLinkModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+	const isValidUrl = () => {
+		try {
+			const parsedUrl = new URL(normalizedUrl);
+			return parsedUrl.protocol === "http:" || parsedUrl.protocol === "https:";
+		} catch {
+			return false;
+		}
+	};
+
+	const isUrlValid = isValidUrl();
```
bggolding, line 32: It's good to see you using variable names starting with `is...` for boolean values.

src/components/GoalList.tsx
```diff
@@ -57,6 +64,20 @@ const GoalList = React.forwardRef<HTMLDivElement, GoalListProps>(({setDraggedIte
             }
         };
 
+        const handleLinkSave = (url: string) => {
+            if (!linkGoal) return;
```
bggolding: How can `linkGoal` be falsy? I don't think this test makes sense.

src/components/GoalList.tsx
```diff
@@ -57,6 +64,20 @@ const GoalList = React.forwardRef<HTMLDivElement, GoalListProps>(({setDraggedIte
...
+
+            dispatch(updateUrlForGoalId({id: linkGoal.id, url}));
+            setLinkGoal(null);
+        };
+
+        const handleLinkRemove = () => {
+            if (!linkGoal) return;
```
bggolding: Ditto

src/components/Tree.tsx
```diff
@@ -359,6 +362,21 @@ const TreeRow = React.forwardRef<HTMLDivElement, TreeRowProps>(({
             </div>
           )}
 
+          {!isEditing && (
```
bggolding, line 365: Please use a 4 space indent. I know that the codebase is inconsistent but we are (slowly) moving to use that.

I also prefer to put the attributes on the same line as the tag to make the code density higher, eg:
```
<div className="link-icon"
     title="Related link"
     aria-label={`Related link for ${treeItem.content || treeItem.type}`}
     onMouseDown={(event) => event.stopPropagation()}
     onClick={(event) => {
         event.stopPropagation();
         onLinkClick(treeItem);
     }}>
     [...]
</div>
```

src/components/types.ts
```diff
@@ -49,6 +49,7 @@ export type ParsedGoalId = ParsedFunctionalId | ParsedNonFunctionalId;
 
 export interface ClusterGoal extends GoalBase {
     SubGoals: ClusterGoal[]
+    url?: string;
```
bggolding: I think it makes more sense to define this as:
```typescript
    url: string | undefined;
```
Ie, the `url` is always there but it can be undefined.

bggolding (changes requested):
I've had a long look at your code and overall it's good. I've made quite a few comments which I'm open to discuss with you further.

src/components/GoalList.tsx
ShengboYY: Yeah I see. The save callback is only available while a goal is selected, so the null check was redundant. I now only show the modal when a goal is selected, and pass that goal directly to the save handler.

src/components/Graphs/GraphHelpers.tsx
ShengboYY: The URL is passed through this flow:
goal.url → description bucket → renderNonFunction → registerGoalLink → click handler

We need to keep it in the bucket because non-functional goals are rendered later. If the rendered symbol represents a single goal, its label becomes clickable. If it represents multiple goals, no link is added because the target would be ambiguous.

src/components/Graphs/GraphHelpers.tsx
ShengboYY: Thanks. I’ve updated the helper to return a new style object and changed both call sites to use the returned value.

src/components/GoalLinkModal.tsx
ShengboYY: I’ve simplified GoalLinkModal so it only accepts goal and onClose. It is now rendered only when a goal is selected, and the save/remove logic is handled inside the modal instead of being duplicated in the parent components.

src/components/Graphs/GraphHelpers.tsx
ShengboYY: Yeah, the original comment was unclear. A goal only has one URL, but a rendered non-functional symbol may combine several goals, each potentially with its own URL. I’ve clarified the comment and wrapped the ternary condition in parentheses. If you’d prefer a different behavior for aggregated symbols, such as showing a list of available links, I’m happy to discuss it.

src/components/types.ts
ShengboYY: Resolved

src/components/Tree.tsx
ShengboYY: Resolved

src/components/GoalList.tsx
ShengboYY: Resolved

src/components/context/treeDataSlice.ts
ShengboYY: Resolved

src/components/context/treeDataSlice.ts
ShengboYY: Resolved

src/components/context/treeDataSlice.ts
ShengboYY: Resolved

src/components/context/treeDataSlice.ts
ShengboYY: Resolved

src/components/Graphs/GraphWorker.tsx
ShengboYY: Resolved

src/components/GoalLinkModal.tsx
ShengboYY: Resolved

src/components/GoalLinkModal.tsx
ShengboYY: Resolved

bggolding:
Could you resolve the conflicts here so I can move towards merging your changes?

PR #298, opened 2026-08-04
Fix image asset paths and improve PNG export quality

Shirley-H99:
## Summary

This PR fixes two issues related to image loading and PNG export quality.

### 1. Fix image loading

Some icons were referenced using inconsistent hard-coded relative paths, such as `img/...` and `./leon.png`. When the application was deployed under the `/mm-local-editor/` base path, these references could resolve to incorrect locations and result in `404` errors.

The image references have been replaced with Vite static imports. The generated asset URLs are now used consistently in both MaxGraph's `imagePath` configuration and React `` elements, ensuring that the icons load correctly in both development and production environments.

### 2. Improve exported PNG quality

Exported PNG files appeared blurry because the graph was rendered using the SVG's on-screen resolution.

The export process now creates a canvas at three times the original dimensions and scales the drawing context accordingly.

### DIFF
--- a/src/components/DetailCarousel.tsx
+++ b/src/components/DetailCarousel.tsx
@@ -1,12 +1,13 @@
 import React from "react";
 
 import Carousel from "react-bootstrap/Carousel";
+import LeonIcon from "/leon.png";
 
 export const DetailCarousel: React.FC = () => (
     <Carousel interval={null}>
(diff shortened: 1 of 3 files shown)


bggolding (approved):
They are very simple and straightforward fixes. Nice work!

Issue #247, opened 2026-02-04
Use regular expression to test and breakdown instanceId


bggolding:
There's a lot of code that's a bit tricky to understand which breaks down `InstanceId`'s. This would be simpler to understand if it was done using a regular expression.

PR #305, opened 2026-08-11
Fix: badly-formatted-instance-id--5-1

### LINKED ISSUE #287: Badly formatted instance ID "-5-1" happened when trying to drag the goal second time

ShengboYY:
## Summary

- Support negative goal IDs in `InstanceId`.
- Replace `split("-")` with regex-based parsing and validation.
- Fix the `badly formatted instanceId "-5-1"` error.

### DIFF
--- a/src/components/utils/GraphUtils.tsx
+++ b/src/components/utils/GraphUtils.tsx
@@ -185,25 +185,24 @@ export function generateCellId<T extends keyof IdsForType>(type: T, ids: IdsForT
     }
 }
 
+// The goal ID may be negative, while the reference suffix remains non-negative.
+const INSTANCE_ID_PATTERN = /^(-?\d+)-(\d+)$/;
+
 export const validateInstanceId = (id: string): InstanceId => {
-    const bits = id.split("-").map(s => s.trim());
-    if (bits.length !== 2) {
+    if (!INSTANCE_ID_PATTERN.test(id)) {
         throw new Error(`badly formatted instanceId "${id}"`);
-    }
-    const [goalId, refId] = bits.map(Number);
-    if (isNaN(goalId) || isNaN(refId)) {
-        throw new Error(`instanceId must contain two numbers, got "${id}"`);
     }
     return id as InstanceId;
 };
 
 export const parseInstanceId = (instanceId: InstanceId) => {
-    const bits = instanceId.split("-").map(s => s.trim());
-    if (bits.length !== 2) {
+    const match = INSTANCE_ID_PATTERN.exec(instanceId);
+    if (!match) {
         throw new Error(`badly formatted instanceId "${instanceId}"`);
     }
 
-    const [goalId, refId] = bits.map(Number);
+    const goalId = Number(match[1]);
+    const refId = Number(match[2]);
 
     return {goalId, refId};
 };


bggolding (changes requested):
We should use a different separator instead of `-` as I described on the ticket #287. It was a poor choice to start with. Personally, I think I would choose `/` or maybe `:`.

bggolding:
Note that this will also close #247

src/components/utils/GraphUtils.tsx
```diff
@@ -185,25 +185,24 @@ export function generateCellId<T extends keyof IdsForType>(type: T, ids: IdsForT
     }
 }
 
+// The goal ID may be negative, while the reference suffix remains non-negative.
+const INSTANCE_ID_PATTERN = /^(-?\d+)-(\d+)$/;
```
bggolding: Instead of `INSTANCE_ID_PATTERN`, please call this `INSTANCE_ID_RE` (for `R`egular `E`xpression). It makes it very clear what it's used for.

src/components/utils/GraphUtils.tsx
```diff
@@ -88,30 +88,26 @@ export function fixEditorPosition(graph: Graph) {
...
 }
 
 export const parseFuncGoalRefId = (id: string): {goalId: TreeGoal["id"], instanceId: InstanceId} => {
-    // Example: Functional-2-1 -> id = "2-1"
-    const parts = id.split("-");
-    if (parts.length !== 2 || parts[0] === "" || parts[1] === "") {
+    // A graph cell keeps its type outside the canonical instance ID, e.g. Functional-2:1.
+    let instanceId: InstanceId;
+    try {
+        instanceId = validateInstanceId(id.trim());
```
bggolding: I am not sure we need this `trim()` here.

src/components/utils/GraphUtils.tsx
```diff
@@ -88,30 +88,26 @@ export function fixEditorPosition(graph: Graph) {
...
-    return `${goal.GoalType}-${goal.instanceId}` as InstanceId;
+    return `${goal.GoalType}-${goal.instanceId}`;
 }
 
 export const parseFuncGoalRefId = (id: string): {goalId: TreeGoal["id"], instanceId: InstanceId} => {
-    // Example: Functional-2-1 -> id = "2-1"
-    const parts = id.split("-");
-    if (parts.length !== 2 || parts[0] === "" || parts[1] === "") {
+    // A graph cell keeps its type outside the canonical instance ID, e.g. Functional-2:1.
+    let instanceId: InstanceId;
```
bggolding: Rather than use `let` here, I would wrap the whole of the function body in `try`/`catch` and use that. The code here is so simple that covering a bit of extra scope doesn't make it less understandable. ie:
```typescript
export const parseFuncGoalRefId = (id: string): {goalId: TreeGoal["id"], instanceId: InstanceId} => {
    try {
        const instanceId = validateInstanceId(id);
        const {goalId} = parseInstanceId(instanceId);

        return {goalId, instanceId};
    } catch {
        throw new Error(`invalid InstanceId: got "${id}"`);
    }
};
```

src/components/utils/GraphUtils.tsx
```diff
@@ -88,30 +88,26 @@ export function fixEditorPosition(graph: Graph) {
...
+    try {
+        instanceId = validateInstanceId(id.trim());
+    } catch {
         throw new Error(`invalid id: got "${id}"`);
```
bggolding: Don't say `invalid id` here, be completely specific, it is an `invalid InstanceId`. That tells the person debugging that it's the rules about InstanceId that aren't being conformed to.

src/components/utils/GraphUtils.tsx
```diff
@@ -185,29 +179,38 @@ export function generateCellId<T extends keyof IdsForType>(type: T, ids: IdsForT
...
+    const match = INSTANCE_ID_PATTERN.exec(instanceId);
+    if (!match) {
         throw new Error(`badly formatted instanceId "${instanceId}"`);
     }
 
-    const [goalId, refId] = bits.map(Number);
+    const goalId = Number(match[1]);
+    const refId = Number(match[2]);
 
     return {goalId, refId};
```
bggolding: I would probably merge these lines into a single statement:
```typescript
return {
    goalId: Number(match[1]),
    refId: Number(match[2])
};
```

src/components/types.ts
```diff
@@ -3,7 +3,16 @@ import {z} from "zod";
...
 // Core types (defined first to avoid circular refs)
 // ============================================
-export type InstanceId = `${number}-${number}`
+export const INSTANCE_ID_SEPARATOR = ":";
+export type InstanceId = `${number}:${number}`
+
+export const createInstanceId = (goalId: number, refId: number): InstanceId => {
+    // Instance IDs are made only from integers, so neither component can contain the separator.
+    if (!Number.isInteger(goalId) || !Number.isInteger(refId) || refId < 0) {
+        throw new Error(`invalid instance ID components: "${goalId}" and "${refId}"`);
```
bggolding: Report these errors separately so that the person debugging doesn't need to look in to the code to work out which one had the error in it.

```typescript
    if (!Number.isInteger(goalId)) {
        throw new Error(`non-numeric goalId: "${goalId}"`);
    }
    if (!Number.isInteger(refId)) {
        throw new Error(`non-numeric refId: "${refId}"`);
    }
    if (refId < 0) {
        throw new Error(`negative refId: "${refId}"`);
    }
```

src/components/types.ts
```diff
@@ -3,7 +3,16 @@ import {z} from "zod";
 // ============================================
 // Core types (defined first to avoid circular refs)
 // ============================================
-export type InstanceId = `${number}-${number}`
+export const INSTANCE_ID_SEPARATOR = ":";
+export type InstanceId = `${number}:${number}`
```
bggolding: This really ought to be:
```typescript
export type InstanceId = `${number}${INSTANCE_ID_SEPARATOR}${number}`
```
That way, if it needs to change again, there are fewer code changes needed.

src/components/utils/GraphUtils.tsx
```diff
@@ -185,29 +179,38 @@ export function generateCellId<T extends keyof IdsForType>(type: T, ids: IdsForT
     }
 }
 
+// New state uses ":"; the legacy pattern is accepted only while stored/imported models are normalised.
+const INSTANCE_ID_PATTERN = /^(-?\d+):(\d+)$/;
```
bggolding: Is there a way to replace the `:` in the reg exp with `INSTANCE_ID_SEPARATOR` so that it's updated automatically if that needs to change again?

src/components/utils/GraphUtils.tsx
```diff
@@ -185,29 +179,38 @@ export function generateCellId<T extends keyof IdsForType>(type: T, ids: IdsForT
...
+const LEGACY_INSTANCE_ID_PATTERN = /^(-?\d+)-(\d+)$/;
```
bggolding: Please call this `LEGACY_INSTANCE_ID_RE`.

src/components/utils/GraphUtils.tsx
bggolding: Putting extra checks like this in the code actually tells the programmer that we aren't sure what the structure of this value is, perhaps it could come from user input. In fact that's not true, these values are internally generated. These little extra bits of information add to the extra cognitive load that the programmer needs to keep in mind when thinking about the program.

bggolding (changes requested):
There are a few things that could be better which I've commented on here.

The testing looks good and appears to catch most cases.

src/components/types.ts
```diff
@@ -4,12 +4,20 @@ import {z} from "zod";
 // Core types (defined first to avoid circular refs)
 // ============================================
 export const INSTANCE_ID_SEPARATOR = ":";
-export type InstanceId = `${number}:${number}`
+export type InstanceId = `${number}${typeof INSTANCE_ID_SEPARATOR}${number}`
```
bggolding: I don't think there should be a `typeof` in this expression, just `${INSTANCE_ID_SEPARATOR}`

src/components/types.ts
```diff
@@ -88,7 +96,7 @@ export const GoalTypeSchema = z.enum(
 );
 
 const instanceIdSchema = z.custom<InstanceId>((val) => {
```
bggolding: Why not make the type of the argument `val` be `string` so that the typing is done at compile time, not at run time? Like this:
```typescript
const instanceIdSchema = z.custom<InstanceId>((val: string) => {
    [...]
};
```

bggolding (changes requested):
The tests that you've added are good and really get into the fine detail of checking the InstanceId's. There are couple of other small issues that could have a little more attention.

bggolding:
You might need to merge in develop to this branch to pick up the fix for the CI issue.

ShengboYY:
@bggolding Hi Ben, I’ve merged develop to pick up the CI fix and resolved the conflict. I also addressed the InstanceId review comments and added regression tests.

src/components/types.ts
```diff
@@ -4,7 +4,8 @@ import {z} from "zod";
 // Core types (defined first to avoid circular refs)
 // ============================================
 export const INSTANCE_ID_SEPARATOR = ":";
-export type InstanceId = `${number}${typeof INSTANCE_ID_SEPARATOR}${number}`
+type InstanceIdSeparator = typeof INSTANCE_ID_SEPARATOR;
+export type InstanceId = `${number}${InstanceIdSeparator}${number}`
```
bggolding: I hadn't quite registered that we were crossing between using a javascript `const` and a typescript `type` and that these worlds don't overlap cleanly. I would go back to how it was and use:
```typescript
export type InstanceId = `${number}${typeof INSTANCE_ID_SEPARATOR}${number}`
```
but maybe add a short comment saying we need to `typeof` to get the value from the `const` to use in a `type`.

bggolding (changes requested):
I'm sorry to have to ask you to change this back to what you'd originally done but it should be easy.

bggolding (approved):
Thanks, that looks good.

PR #309, opened 2026-08-14
Fix: prevent duplicate nested goals

### LINKED ISSUE #306: Duplicate check ignores goals nested below the top level
## Description

When adding a goal to the hierarchy, the duplicate check only scans top-level goals. If an existing goal is moved under a parent, it is no longer detected and can be added again.

ShengboYY:
## Summary

- Use `treeIds` instead of scanning only top-level goals.
- Prevent nested goals from being added again.
- Apply the same duplicate check to single and group additions.


bggolding (approved):
This looks good really. I like how you've updated and clarified the comment with the new way that you're analysing the tree.

While I don't think you've done this here, be aware of not commenting about how the code used to be and how you've changed it. Instead, just try to focus on describing what this code does.

